Issue #23, opened 2021-09-10
Improve loop exit behavior for deprovision roks


mdickens:
Sometimes Even after a cluster is deprovisioned, the loop [here](https://github.com/ibm-mas/ansible-devops/blob/a83974afe3e283a65a665a8506bb2b504aaf41a2/ibm/mas_devops/roles/ocp_deprovision/tasks/providers/roks.yml#L56) is not exiting. Will add an additional conditional check to try to get the loop to exit faster.

```
    "rc": 0,

    "retries": 31,

    "start": "2021-09-09 20:10:38.036981",

    "stderr": "",

    "stderr_lines": [],

    "stdout": "0\nRetrieving cluster fvtrelease...",

    "stdout_lines": [

        "0",

        "Retrieving cluster fvtrelease..."

    ]
```

mdickens:
https://github.com/ibm-mas/ansible-devops/pull/22

mdickens:
Fixed with https://github.com/ibm-mas/ansible-devops/pull/22

mdickens:
thinking some more im not sure this fix is done... when it gets to the next step for provision , if it is still returning 0 it possibly may not even try to create the cluster at that step...

Either way i reopened the issue above because now not it is not waiting enough time ... It looks like it exited out of the first loop because it found that output of exactly 2 lines:

So it didnt wait at all for the deprovision to take effect... Ill add a minimum number of attempts before the new criteria will allow the check to break the loop for this problem.

PR #25, opened 2021-09-10
Update roks.yml to wait for deprovision as now it is not waiting long enough.

mdickens:
Update roks.yml to require a minimum 5 attempts during the Wait for cluster to be deprovisioned task in order for the condition checking for 2 lines of stdout to be valid.

https://github.com/ibm-mas/ansible-devops/issues/23 .


durera:
See PR#24

PR #43, opened 2021-10-21
[patch] - install BAS on ROKS

andrercm:


docs/playbooks/bas.md
```diff
@@ -1,20 +1,20 @@
 # BAS Playbooks
 
 ## Install BAS
```
stonepd, line 3: Could you add a description of what the playbook actually does, maybe a summary of the README.md: `Installs **IBM Behavior Analytics Services** on IBM Cloud Openshift Clusters (ROKS) and generates configuration that can be directly applied to IBM Maximo Application Suite.`

ibm/mas_devops/roles/bas_install/templates/bascfg.yml.j2
```diff
@@ -30,6 +30,42 @@ spec:
...
       crt: |
-        {{bas_tls_crt | indent(8) }}
\ No newline at end of file
+        {{bas_tls_crt[0] | indent(8) }}
+    - alias: part2
+      crt: |
+        {{bas_tls_crt[1] | indent(8) }}
+    - alias: part3
+      crt: |
+        -----BEGIN CERTIFICATE-----
```
stonepd, line 41: Is it OK to include this certificate in our code? I see that is has come from an existing public source: [isrgrootx1.pem](https://letsencrypt.org/certs/isrgrootx1.pem) but can you confirm that we are able to include this in our code or should we dynamically include it at script run-time?

stonepd (changes requested):
Technically it looks good, can you address the review comments please

andrercm:
> I want to make sure its ok to include the tls cert in the public github. Is it ok?

@racree i would think it should be ok as the cert i added to make it work is publicly available:
https://letsencrypt.org/certs/isrgrootx1.pem

ibm/mas_devops/roles/bas_install/templates/bascfg.yml.j2
andrercm: i would think it should be okay as it is publicly available: https://letsencrypt.org/certs/isrgrootx1.pem
whether we want to hardcode it or include it via another way... I don't know.
@durera @CAIOFCP any concerns with this approach?

andrercm:
@stonepd responded your points, let me know if you think we should change something else

stonepd (approved):

PR #54, opened 2021-11-13
[patch] externalize Service Binding Operator subscription settings

terenceq:
Currently the Service Binding Operator subscription is hard coded to the preview channel with a starting version of 0.8 and and update strategy of manual. As Maximo Application Suite matures the need to move away from Service Binding Operator v0.8 will be required. 

This PR externalizes three variables from the `oc_setup_mas_deps`role:

- sbo_channel
- sbo_startingcsv
- sbo_plan_approval

The defaults for these new variables will remain:

```yaml
sbo_channel: preview
sbo_startingcsv: service-binding-operator.v0.8.0
sbo_plan_approval: Manual
```

The configure-ocp playbook allows them to be overridden by checking for the following environment variables at runtime:

SBO_CHANNEL
SBO_STARTINGCSV
SBO_PLAN_APPROVAL


stonepd (approved):
Looks good to me

PR #57, opened 2021-11-17
[patch] fix sls secret registrationKey

racree:
Update the {{sls_instance_name}}-bootstrap secret's key name "registration_key" to "registrationKey"

### DIFF
--- a/ibm/mas_devops/roles/sls_install/tasks/bootstrap.yml
+++ b/ibm/mas_devops/roles/sls_install/tasks/bootstrap.yml
@@ -11,4 +11,4 @@
       stringData:
         licensingId: "{{ (bootstrap.license_id is defined or bootstrap.license_id is defined == '') | ternary(bootstrap.license_id, omit ) }}"
         licensingKey: "{{  (bootstrap.entitlement_file is defined or bootstrap.entitlement_file is defined == '') | ternary(lookup('file', bootstrap.entitlement_file), omit) }}"
-        registration_key: "{{ (bootstrap.registration_key is defined  or bootstrap.registration_key == '') | ternary(bootstrap.registration_key, omit) }}"
+        registrationKey: "{{ (bootstrap.registration_key is defined  or bootstrap.registration_key == '') | ternary(bootstrap.registration_key, omit) }}"


twhart (approved):

ibm/mas_devops/roles/sls_install/tasks/bootstrap.yml
```diff
@@ -11,4 +11,4 @@
       stringData:
         licensingId: "{{ (bootstrap.license_id is defined or bootstrap.license_id is defined == '') | ternary(bootstrap.license_id, omit ) }}"
         licensingKey: "{{  (bootstrap.entitlement_file is defined or bootstrap.entitlement_file is defined == '') | ternary(lookup('file', bootstrap.entitlement_file), omit) }}"
-        registration_key: "{{ (bootstrap.registration_key is defined  or bootstrap.registration_key == '') | ternary(bootstrap.registration_key, omit) }}"
+        registrationKey: "{{ (bootstrap.registration_key is defined  or bootstrap.registration_key == '') | ternary(bootstrap.registration_key, omit) }}"
```
alequint, line 14: @racree this is something I have been discussing for some time already with @CAIOFCP in the past, but we don't really  need registration_key to bootstrap SLS. Registration key is created when we deploy LicenseService operand. So that automation obtain it from there and use in SLSCfg... I am wondering if you had any problem to configure SLS by not informing registration key here (?)

alequint: @twhart ⬆️

twhart: The problem is this is a bug, we already set this during bootstrap but it's the wrong key value

twhart: For more info, the registration key is not required but does simplify things if you're restoring from a backup. For example if MAS is using SLS with a specific registration key then all you would need to do is reregister without updating your slscfg. But right now the scripts will not allow you to set the registration key because the key "registration_key" is wrong, SLS will read the secret and only recognize the key "registrationKey" so this needs to be update if it's going to be used at all.

PR #62, opened 2021-11-22
[patch] adding default storage class for bas

andrercm:
Thanks to Hari Palleti, realized I wasnt setting BAS storage classes as default for other common playbooks we use in our collection, so just fixing it so it installs BAS instance even if you don't set any BAS variable.

Same to what we have here: https://github.com/ibm-mas/ansible-devops/blob/master/ibm/mas_devops/playbooks/bas/install-bas.yml#L7

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -5,14 +5,14 @@
 ```bash
 cd ibm/mas_devops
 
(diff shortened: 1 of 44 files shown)


durera:
PR is no longer required, issues has been addressed in bugfixes2411 branch

Issue #15, opened 2021-09-02
Add must-gather role to collection


durera:
We should be able to execute must-gather in a cluster by running a simple role: `ibm.mas_devops.must_gather`

It's really simple to run must-gather for MAS:
```
oc adm must-gather --image=quay.io/aiasupport/must-gather -- gather -cgl --noclusterinfo --mas-instance-id {{mas_instance_id}}
```

- The role should obviously support passing in the mas instance id
- The role should support configuring where must gather output is saved locally
- The role should execute must gather, and once complete report where the output from the gather can be located on the users local system.
- Obviously, we need new documentation added
- Update all the (internal) MAS FVT scenarios to execute must-gather after testing is complete, and validate that it contains the expected resources, so that our FVT is also validating must-gather processes still work

Reference: https://docs.openshift.com/container-platform/4.6/support/gathering-cluster-data.html#about-must-gather_gathering-cluster-data

PR #91, opened 2021-12-06
[minor] Add support for AI Applications' must-gather tooling

caroazad:
I have included a playbook and a role for the must gather as in issue #15 
The must gather requires a base directory on the local machine where the results of the must-gather would be stored. The directory must be a full path. 
Other needed variable is the mas instance variable.

To test, run oc login for ibm cloud before running the run_must_gather.yml playbook


durera:
Works a treat ...

image

durera (dismissed):

PR #93, opened 2021-12-08
[minor] Add support for CP4D v4

andrercm:
Add support for CloudPak for Data (CP4D) v4.  This support is in addition to the existing support for CP4D v3.5, which will remain unaffected by this update.

Where possible we will attempt to intelligently determine whether CP4D v4 or v3.5 should be installed, but the user can also directly control which version is used if they choose.

This update also affects cert-manager installation, as CP4D will auto-install a version of cert-manager that is incompatible with the version of cert-manager used by MAS prior to v8.7.  The roles and playbooks will attempt to handle all this complexity for the user automatically.

## Final test scenarios to get this released:
- [x] lite-manage-roks playbook (configured for cpd v3.5 / ocp 4.6 / mas 8.6.x channel)
- [x] lite-manage-roks playbook (configured for cpd v4 / ocp 4.8 / mas 8.7 pre-release build)
- [x] lite-core-roks playbook (configured for cpd v3.5 / ocp 4.6 / mas 8.x channel)
- [ ] sample tekton pipeline (configured for cpd v4 / ocp 4.8 / mas 8.7 pre-release build)
- [ ] sample tekton pipeline (configured for cpd v3.5 / ocp 4.6 / mas 8.x channel)


.github/workflows/collection.yml
```diff
@@ -0,0 +1,47 @@
+name: Build Ansible Collection
```
durera, line 1: This file should not be added back, bad merge.

ibm/mas_devops/playbooks/cp4d/create-db2-instance.yml
```diff
@@ -25,6 +28,16 @@
     db2wh_backup_storage_class: "{{ lookup('env', 'DB2WH_BACKUP_STORAGE_CLASS') }}"
     db2wh_backup_storage_size_gb: "{{ lookup('env', 'DB2WH_BACKUP_STORAGE_SIZE_GB') | default(100, true) }}"
 
+    # Configure CP4D transaction logs storage .. despite the name, this ONLY applies to Db2 in CP4D 4.0
```
durera, line 31: This TODO comment can be removed, it's already been done

ibm/mas_devops/playbooks/cp4d/create-db2-instance.yml
```diff
@@ -25,6 +28,16 @@
...
+    # TODO: rename this to reflect that it's only affecting db2
+    db2wh_logs_storage_class: "{{ lookup('env', 'DB2WH_LOGS_STORAGE_CLASS') }}"
+    db2wh_logs_storage_size_gb: "{{ lookup('env', 'DB2WH_LOGS_STORAGE_SIZE_GB') | default(100, true) }}"
+
+    # Configure CP4D temporary storage .. despite the name, this ONLY applies to Db2 in CP4D 4.0
```
durera, line 36: This TODO comment can be removed, it's already been done

ibm/mas_devops/playbooks/cp4d/create-db2-instance.yml
```diff
@@ -25,6 +28,16 @@
...
+    # TODO: rename this to reflect that it's only affecting db2
+    db2wh_temp_storage_class: "{{ lookup('env', 'DB2WH_TEMP_STORAGE_CLASS') }}"
+    db2wh_temp_storage_size_gb: "{{ lookup('env', 'DB2WH_TEMP_STORAGE_SIZE_GB') | default(100, true) }}"
+
     # TODO: This username is actually the username of Db2, not CP4D
```
durera, line 41: This TODO comment can be removed, it's already been done

ibm/mas_devops/playbooks/cp4d/install-services-db2.yml
```diff
@@ -2,16 +2,24 @@
 - hosts: localhost
   any_errors_fatal: true
   vars:
+    # Cluster details
+    # ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
+    cluster_name: "{{ lookup('env', 'CLUSTER_NAME') }}"
```
durera, line 7: Why do we need the CLUSTER_NAME and CLUSTER_TYPE for installing db2 service?

ibm/mas_devops/playbooks/cp4d/install-services-fullstack.yml
```diff
@@ -2,8 +2,14 @@
 - hosts: localhost
   any_errors_fatal: true
   vars:
+    # Cluster details
+    # ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------
+    cluster_name: "{{ lookup('env', 'CLUSTER_NAME') }}"
```
durera, line 7: Why do we need the CLUSTER_NAME and CLUSTER_TYPE for installing cpd services?

ibm/mas_devops/playbooks/cp4d/install-services-watsonstudio.yml
```diff
@@ -2,12 +2,20 @@
 - hosts: localhost
   any_errors_fatal: true
   vars:
+    # Cluster details
```
durera, line 5: Why do we need the CLUSTER_NAME and CLUSTER_TYPE for installing cpd services?

ibm/mas_devops/playbooks/ocp/configure-ocp.yml
```diff
@@ -2,6 +2,7 @@
 - hosts: localhost
   any_errors_fatal: true
   vars:
+    mas_channel: "{{ lookup('env', 'MAS_CHANNEL') | default('8.x', True)}}"
```
durera, line 5: Could we put a comment here explaining why the mas channel is needed .. I assume it's to determine how we configure the cluster, e.g. the comment should point the reader to where we can find details of how this is used so it can be easily found.

ibm/mas_devops/roles/cp4d_db2wh/defaults/cpd35/main.yml
```diff
@@ -1,5 +1,6 @@
 # A temporary workaround for a bug in CP4D DB2, details in story 1372 & 1472, to be removed once the bug is fixed in CP4D (4.x version)
 apply_db2u_sts_patch: true
+db2wh_load_from_config: true
```
durera, line 3: unless it's been reimplemented `db2wh_load_from_config` was removed from use in the master branch, it's the only way it works now, and this is unnecessarily added back I think as part of a bad merge?

ibm/mas_devops/roles/cp4d_db2wh/defaults/cpd40/main.yml
```diff
@@ -0,0 +1,11 @@
+# A temporary workaround for a bug in CP4D DB2, details in story 1372 & 1472, to be removed once the bug is fixed in CP4D (4.x version)
+apply_db2u_sts_patch: false
+db2wh_load_from_config: false
```
durera, line 3: unless it's been reimplemented `db2wh_load_from_config` was removed from use in the master branch, it's the only way it works now, and this is unnecessarily added back I think as part of a bad merge?

ibm/mas_devops/roles/cp4d_db2wh_manage_hack/tasks/main.yml
```diff
@@ -26,10 +26,12 @@
     name: "{{ cp4d_db_config_configmap_name }}-{{ db2wh_instance_name }}"
     namespace: "{{ cpd_meta_namespace }}"
   register: cp4d_db_config_info
+  when: db2wh_load_from_config == true
```
durera, line 29: `db2wh_load_from_config` was eliminated in master to streamline things ... can we do the same here, there's no need for a load from config == false mode as there is only one way to provision now (cp4d_db2wh_api replaced cp4d_db2wh instead of being an alternative role)

ibm/mas_devops/roles/cp4d_db2wh_manage_hack/tasks/main.yml
```diff
@@ -26,10 +26,12 @@
...
 
 - name: Read the DB2 database ID for {{ db2wh_instance_name }} from DB config ConfigMap data
   set_fact:
     db2wh_instance_id: "db2wh-{{ cp4d_db_config_info.resources[0].data.cp4d_db_id }}"
+  when: db2wh_load_from_config == true
```
durera, line 34: `db2wh_load_from_config` was eliminated in master to streamline things ... can we do the same here, there's no need for a load from config == false mode as there is only one way to provision now (cp4d_db2wh_api replaced cp4d_db2wh instead of being an alternative role)

ibm/mas_devops/roles/cp4d_db2wh_manage_hack/tasks/main.yml
```diff
@@ -82,6 +84,7 @@
 #   LOAD,CREATE_EXTERNAL_ROUTINE,QUIESCE_CONNECT,SECADM ON DATABASE TO USER {{ db2wh_username }}
 
 - name: Prepare db2 for use with Maximo TPAE
+  ignore_errors: yes
```
durera, line 87: Can we have a comment here explaining why it's okay to ignore errors please, for the benefit of our future selves?

ibm/mas_devops/roles/cp4d_install/tasks/main.yml
```diff
@@ -1,69 +1,64 @@
...
 
-# 3. Install CP4D operator
+# 4. Check if CPD version is compatible with cluster type
 # -----------------------------------------------------------------------------
-- name: "Install CPD Subscription"
-  community.kubernetes.k8s:
-    apply: yes
-    definition: "{{ lookup('template', 'templates/subscription.yaml') }}"
-  register: _cpd_subscription_result
+# - fail if cluster_type = fyre/quickburn and cpd_version = cpd40
```
durera, line 51: Can we take this out please .. let's not codify current limitations based on cluster type .. there's no reason the infrastructure provider should impact this ... and presumably this answers my earlier question about why the cluster_type and cluster_name had leaked into something that shouldn't be dealing with such things.

ibm/mas_devops/roles/ocp_setup_mas_deps/tasks/development-catalogs.yml
```diff
@@ -23,17 +23,27 @@
         .dockerconfigjson: "{{ content | join('') | string }}"
   register: result
 
-# 2. Patch the default service account in openshift-marketplace so that it is able to pull from Artifactory
-- name: "Patch default SA on openshift-marketplace"
+# 2. Crete the service accounts in openshift-marketplace so that it is able to pull from Artifactory
```
durera, line 26: Small typo: `Crete`

durera: Also, jsut for my education really, why do we need all the extra service accounts now?

ibm/mas_devops/roles/ocp_setup_mas_deps/tasks/main.yml
```diff
@@ -28,13 +28,15 @@
...
 # TODO: Version check existing installation, ensure at least 1.2 is installed
 #       Fail if version < 1.2 is installed
-
 - name: Deploy Cert Manager
+  vars:
+    mas87_channels: "{{ mas87_default_channels }}"
   shell: |
     oc create namespace cert-manager
-    oc apply -f https://github.com/jetstack/cert-manager/releases/download/v1.2.0/cert-manager.yaml
+    oc apply -f https://github.com/jetstack/cert-manager/releases/download/v1.6.1/cert-manager.yaml
```
durera, line 36: where's the code that will install cert-manager 1.2 when MAS <8.7?  Am I missing something here, this looks like this will only work for MAS 8.7 + now, and we do nothing in the case of earlier versions ... we want the playbook to continue to install cert-manager 1.2 for MAS 8.6 (or less) installs as we don't want the hassle/don't have time to re-certify/re-test all prior releases against 1.6.

yamllint.yaml
```diff
@@ -1,3 +1,7 @@
 extends: relaxed
 rules:
   line-length: disable
+
```
durera, line 4: this is a bad merge too, this doesn't work in the config file and had to be added as a command line flag (elsewhere)

durera (changes requested):
See individual comments 👍

ibm/mas_devops/playbooks/cp4d/install-services-db2.yml
andrercm: We can probably remove name, but I'm using 'cluster_type' for now as a check to verify whether it's a roks cluster or not. CPD v4 install has specific roks steps to reload worker nodes that won't work for non-roks clusters

.github/workflows/collection.yml
andrercm: Fixed.

ibm/mas_devops/playbooks/cp4d/create-db2-instance.yml
andrercm: Fixed.

ibm/mas_devops/playbooks/cp4d/create-db2-instance.yml
andrercm: Fixed.

ibm/mas_devops/playbooks/cp4d/create-db2-instance.yml
andrercm: Fixed.

ibm/mas_devops/roles/cp4d_db2wh_manage_hack/tasks/main.yml
andrercm: Fixed. Reason why i added it is because the same shell script returns a successful output for db2wh in cpd 3.5 but returns a fail for cpd 4.0 as it complains about granting privileges to self (db2inst1), although the overall db setup was successful, and manage/health can be installed just fine.

ibm/mas_devops/roles/cp4d_install/tasks/main.yml
andrercm: Yes, that's right - as I mentioned above, we're limited for now to install cpd v4 with this code on ROKS.
I can remove it, but if someone tries to run this against a non-roks, then 'cpd_install' role will fail here:
https://github.com/ibm-mas/ansible-devops/blob/manage-components-cpd40/ibm/mas_devops/roles/cp4d_install/tasks/cpd40/cpd-install.yml#L42

Should I simply remove this validation/check? Any suggestions?

ibm/mas_devops/roles/ocp_setup_mas_deps/tasks/main.yml
andrercm: True, good point - Added 1.6.1 as it's the supported by OCP 4.8, it seems then we'd need some logic to determine which version to install accordingly to the mas_channel as well?
i.e if not a MAS 8.7 then go with 1.2 version.

Problem is... if it is a MAS 8.7, we might have two options: with CPD 4.0 (i.e if using `lite-manage-roks.yml`) or without CPD 4.0 (i.e using `lite-core-roks.yml`

Any suggestions? 
I could simply rollback this to 1.2 if we assume MAS 8.7 will always use cert-manager that comes with CPD v4's...

yamllint.yaml
andrercm: Sorry, rolled back from master to fixed it

ibm/mas_devops/roles/ocp_setup_mas_deps/tasks/development-catalogs.yml
andrercm: 1# - Fixed typo. 
2# - Discussed this with @CAIOFCP it seems this is really due a change on OCP 4.8, each catalog source operator now relies on its own SA for the `imagePullSecret` as in:
```
spec:
  nodeSelector:
    kubernetes.io/os: linux
  restartPolicy: Always
  serviceAccountName: ibm-mas-iot-operators
  imagePullSecrets:
    - name: wiotp-docker-local
    - name: ibm-mas-iot-operators-dockercfg-cb7zv
```
So, we just need to create the corresponding SA and patch it with the `wiotp-docker-local` so it authorizes the image pull successfully... haven't removed the `default` SA so it keeps compatibility with OCP 4.6 way.

ibm/mas_devops/roles/ocp_setup_mas_deps/tasks/main.yml
andrercm: I can control which version of cert-manager to install based on ocp_version if that's ok?

If mas_channel is not a MAS 8.7, then jetstack cert-manager will be installed, but then we do a check based on the ocp_version, if this is an ocp 4.6 = install cm v1.2 otherwise install v1.6.1?
```
- name: Deploy Cert Manager
  vars:
    mas87_channels: "{{ mas87_default_channels }}"
    regex: '\d\.\d'
    os_version: "{{ ocp_version | regex_findall(regex) | join(',') }}"
    cm_version: "{{ 'v1.2' if (os_version == '4.6') else 'v1.6.1' }}" #if we have to install jetstack cert-manager, need to check what's the ocp version that is compatible
  shell: |
    oc create namespace cert-manager
    oc apply -f https://github.com/jetstack/cert-manager/releases/download/{{ cm_version }}/cert-manager.yaml
  when:
    - (_cert_manager_deployed.resources | length) == 0
    - mas_channel not in mas87_channels #Don't install this cert-manager if MAS 8.7 is being installed
    - not airgap_install
```

andrercm: if MAS 8.7 is set along with playbooks we still use mas_channel to determine the installation of cpd v4, and it will skip cert-manager standalone installation.

ibm/mas_devops/playbooks/cp4d/create-db2-instance.yml
durera: reviewed

ibm/mas_devops/playbooks/cp4d/install-services-db2.yml
durera: > CPD v4 install has specific roks steps to reload worker nodes after we setup image pull secret that won't work for non-roks clusters

That sounds horrible, that can't be their normal install process surely!!  But that's a question/topic for another day to take up with the CP4D team.

ibm/mas_devops/roles/cp4d_db2wh_manage_hack/tasks/main.yml
durera: Do we have an action item to address this oddity/make it pass even in cpd v4?

ibm/mas_devops/roles/ocp_setup_mas_deps/tasks/development-catalogs.yml
durera: Thanks, I also noticed that we don't have all the app catalogs covered .. so I added a future TODO, but we don't care for this release:

```
  with_items:
    - default
    - ibm-mas-operators
    - ibm-truststore-mgr-operators
    - ibm-mas-manage-operators
    - ibm-mas-iot-operators
    - ibm-sls-operators
    - ibm-data-dictionary-operators
    # TODO: we are missing other application catalogs here!
```

ibm/mas_devops/roles/ocp_setup_mas_deps/tasks/main.yml
durera: We need to get on a call me, you and Joao to go over what myself and Joao have previously discussed about how cert-manager will be managed in 8.7 + 👍

andrercm: Added a piece of code to install foundation services + ibm-cert-manager operandRequest if we identify the need to install cert-manager in case of a MAS 8.7 channel being installed with the playbooks. 
Could you please review?

ibm/mas_devops/roles/cp4d_db2wh_manage_hack/tasks/main.yml
andrercm: Let me test it out some more and see if I can get it removed, the reason why I added it in first place was because db2wh in v4 is installed via db2ucluster currently and it does not require a config-map to store the information for it, and it was failing on the manage-hack role as I don't have a configmap for this version.
So added this check to just use the configmap for db2wh v3.5

ibm/mas_devops/roles/cp4d_db2wh/defaults/cpd40/main.yml
andrercm: Actually while Alex was making the changes for the rest api to work for cpd 3.5, I was working in parallel with the cpd 4.0 and used the db2ucluster approach as base, so to make it work both ways, had to reintroduce this flag back, as db2ucluster approach does not use config map to store db2 info. It's going to take up more time to understand the api changes for v4 to implement same approach as we did for v3.5 db2wh, so maybe this could be an enhancement for near future? Let me know your thoughts.

ibm/mas_devops/playbooks/ocp/configure-ocp.yml
andrercm: Fixed.

ibm/mas_devops/roles/ocp_setup_mas_deps/tasks/main.yml
alequint: Several possibilities :-)

Let me know if the following is aligned with what you guys have in mind:

* OCP 4.6 (MAS 8.6) + CP4D 3.5 (or no CP4D) -> cfg-ocp install cert-manager 1.2
* OCP 4.6 (MAS 8.6) + CP4D 4.0 -> cfg-ocp does nothing -> install-cp4d install cert-manager 1.6 (?) 
* OCP 4.8 (MAS 8.7) + no CP4D -> cfg-cop install cert-manager 1.6
* OCP 4.8 (MAS 8.7) + CP4D 4.0 -> cfg-ocp does nothing -> install-cp4d install cert-manager 1.6

ibm/mas_devops/roles/cp4d_install/tasks/main.yml
andrercm: Exactly, i was surprised when I saw that in the steps... ok, will keep it then.
Here are the cpd40 docs explaining this process if you're interested: 
https://www.ibm.com/docs/en/cloud-paks/cp-data/4.0?topic=tasks-configuring-your-cluster-pull-images

ibm/mas_devops/playbooks/cp4d/install-services-fullstack.yml
andrercm: Fixed. Removed CLUSTER_NAME, not really needed. Maintained CLUSTER_TYPE due ROKS limitation check i.e reload nodes.

ibm/mas_devops/playbooks/cp4d/install-services-watsonstudio.yml
andrercm: Fixed. Removed CLUSTER_NAME, not really needed. Maintained CLUSTER_TYPE due ROKS limitation check i.e reload nodes.

ibm/mas_devops/roles/ocp_setup_mas_deps/tasks/main.yml
andrercm: @alequint we had a call earlier today, and things got simpler than that. Basically `ocp_setup_mas_deps` role will install either foundation services ibm-cert-manager (which is the same one cpd v4 and MAS 8.7 will use) or will continue to install standalone/jetstack cert-manager if MAS channel **is not one** of the 8.7 channels (currently these are 'stable' and 'm2dev87'.) 
So in your 4 scenarios:
OCP 4.6 (MAS 8.6) + CP4D 3.5 (or no CP4D) -> cfg-ocp install cert-manager 1.2 - **MAS 8.6, so it will install cert-manager v1.2.0**
OCP 4.6 (MAS 8.6) + CP4D 4.0 -> cfg-ocp does nothing -> install-cp4d install cert-manager 1.6 (?) - **MAS 8.6, so it will install cert-manager v1.2.0, playbook will fail by the time it tries to install cpd v4 as MAS 8.6 is not compatible with it**
OCP 4.8 (MAS 8.7) + no CP4D -> cfg-cop install cert-manager 1.6 - **MAS 8.7, so it will install foundation services ibm-cert-manager**
OCP 4.8 (MAS 8.7) + CP4D 4.0 -> cfg-ocp does nothing -> install-cp4d install cert-manager 1.6 - **MAS 8.7, so it will install foundation services ibm-cert-manager, before cpd v4 do so which should be okay**

ibm/mas_devops/roles/cp4d_db2wh/defaults/cpd35/main.yml
andrercm: Actually while Alex was making the changes for the rest api to work for cpd 3.5, I was working in parallel with the cpd 4.0 and used the db2ucluster approach as base, so to make it work both ways, had to reintroduce this flag back, as db2ucluster approach does not use config map to store db2 info. It's going to take up more time to understand the api changes for v4 to implement same approach as we did for v3.5 db2wh, so maybe this could be an enhancement for near future? Let me know your thoughts.

ibm/mas_devops/roles/cp4d_db2wh_manage_hack/tasks/main.yml
andrercm: Let me test it out some more and see if I can get it removed, the reason why I added it in first place was because db2wh in v4 is installed via db2ucluster currently and it does not require a config-map to store the information for it, and it was failing on the manage-hack role as I don't have a configmap for this version.
So added this check to just use the configmap for db2wh v3.5

ibm/mas_devops/roles/ocp_setup_mas_deps/tasks/main.yml
alequint: @andrercm for the latest scenario, isn't there a chance to end up with two cert-managers (same version but duplicated installation - assuming they might be installed in different projects?). Or we are installing in the same project cp4d will do?

andrercm: @alequint don't think so because we'll request same ibm-cert-manager that foundation services provides, it's the same one that cpd v4 will request, we're just antecipating this step for this scenario. It will also be installed in ibm-common-services namespace.

alequint: @andrercm here is a problem I noticed during my tests in this sequence of tasks:

```
- name: Check if Cert Manager is already installed
  community.kubernetes.k8s_info:
    api_version: v1
    name: cert-manager
    namespace: "cert-manager"
    kind: Deployment
  register: _cert_manager_deployed

- name: Define which Cert Manager to install
  set_fact:
    cm_version: "{{ 'foundation-services' if (mas_channel in mas87_channels) else 'standalone' }}" # install foundation services cert-manager if MAS 8.7+, otherwise, install jetstack/standalone version
  when:
    - (_cert_manager_deployed.resources | length) == 0
    - not airgap_install

- include_tasks: "tasks/certmanager/{{ cm_version }}.yml"
```

Let's say my cluster has cert-manager deployment already. This [criteria](https://github.com/ibm-mas/ansible-devops/blob/manage-components-cpd40/ibm/mas_devops/roles/ocp_setup_mas_deps/tasks/main.yml#L33) will fail, and so `cm_version` will not have any result. The following [task](https://github.com/ibm-mas/ansible-devops/blob/manage-components-cpd40/ibm/mas_devops/roles/ocp_setup_mas_deps/tasks/main.yml#L36), however, does not have same criteria, and it will have a runtime error.

Maybe it is just a matter of adding the criteria in the last task... however, there is not guarantee if the version already installed will support or not what automation is installing.



Is not used here as well

alequint (changes requested):
Testing pipeline containing latest code present in this branch, BAS cannot be installed after install-amq-streams is due to a conflict between AMQ Streams versions

image

ibm/mas_devops/roles/ocp_setup_mas_deps/tasks/main.yml
andrercm: ah yeah, you're right Alex, we can add same criteria in the include tasks in the bottom as safeguard in this case, just fixed it.
I think at this point we can just control whether it is installed or not, and if not then we install the compatible version with the "to-be" installed mas_channel.
If we need later to ensure cert-manager compatibility, then we would need to add more complexity to check the version and allow or not the mas install.

ibm/mas_devops/playbooks/cp4d/install-services-watsonstudio.yml
andrercm: Actually, figured out I need CLUSTER_NAME, so the reload worker nodes process works.

ibm/mas_devops/roles/cp4d_db2wh/defaults/cpd35/main.yml
andrercm: remove the need of `db2wh_load_from_config` and now if `cpd_version == 'cpd40'` then we don't load the config map, if `cpd_version == 'cpd35'` then we load the config map as rest api expects.

ibm/mas_devops/roles/cp4d_db2wh/defaults/cpd40/main.yml
andrercm: remove the need of `db2wh_load_from_config` and now if `cpd_version == 'cpd40'` then we don't load the config map, if `cpd_version == 'cpd35'` then we load the config map as rest api expects.

ibm/mas_devops/roles/cp4d_db2wh_manage_hack/tasks/main.yml
andrercm: remove the need of `db2wh_load_from_config` and now if `cpd_version == 'cpd40'` then we don't load the config map, if `cpd_version == 'cpd35'` then we load the config map as rest api expects.

ibm/mas_devops/roles/cp4d_db2wh_manage_hack/tasks/main.yml
andrercm: remove the need of `db2wh_load_from_config` and now if `cpd_version == 'cpd40'` then we don't load the config map, if `cpd_version == 'cpd35'` then we load the config map as rest api expects.

alequint (approved):

ibm/mas_devops/roles/ocp_setup_mas_deps/tasks/main.yml
durera: Sounds like this is resolved, marking the conversation resolved.

ibm/mas_devops/playbooks/fullstack-roks.yml
```diff
@@ -87,15 +82,13 @@
     db2wh_instance_name: db2w-shared
     db2wh_load_from_config: "true"
```
durera, line 83: This should be removed, it's no longer used in the role

pipelines/tasks/mas/hack-manage-db2.yaml
```diff
@@ -13,6 +13,10 @@ spec:
       type: string
       description: Name (specifically, not the ID) of the DB2 Warehouse instance to execute the hack
       default: "db2wh-db01"
+    - name: db2wh_load_from_config
```
durera, line 16: This should be removed, it's no longer used in the role (and further below where it's loaded into the stepTemplate too).

docs/playbooks/mas.md
```diff
@@ -152,8 +152,8 @@ This should should be part of the Manage operator, but is not so we have to do i
 The parameters are all optional:
 
 - `CPD_NAMESPACE` namespace where Cloud Pak for Data is installed. Default is `cpd_meta_namespace`
-- `CPD_DB2WH_INSTANCE_NAME` name of the DB2 Warehouse instance created in CP4D. Default is `db2wh-db01` to be compatible with `install-db2` playbook
-- `CPD_DB2WH_db2wh_load_from_config` when `true`, it tells the playbook to obtain DB2 Warehouse instance from the internal config-map created by `install-db2-api` instead of using `CPD_DB2WH_INSTANCE_NAME`. This option must be used when `hack-manage-db2` is executed after `install-db2-api`. Default is `false`
+- `DB2WH_INSTANCE_NAME` name of the DB2 Warehouse instance created in CP4D.
+- `DB2WH_LOAD_FROM_CONFIG` when `true`, it tells the playbook to obtain DB2 Warehouse instance from the internal config-map created by `create-db2-instance` instead of using `DB2WH_INSTANCE_NAME`. This flag must be set `true` when `hack-manage-db2` and, previously, `create-db2-instance`, is used to create instance in CP4D 3.5. Default is `false`, used for databases created in CP4D 4.0.
```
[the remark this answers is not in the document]
andrercm: Fixed, also adjusted the part of the docs where it references old named playbooks `install-db2.yml` and `install-db2-api` to `install-services-db2` and `create-db2-instance.yml` respectively.

ibm/mas_devops/playbooks/fullstack-roks.yml
andrercm: Fixed.

ibm/mas_devops/playbooks/mas/hack-manage-db2.yml
```diff
@@ -5,6 +5,6 @@
     # Ensure these are kept in sync with the values used in cp4d/install-db2.yml
     cpd_meta_namespace: "{{ lookup('env', 'CPD_NAMESPACE') | default('cpd-meta-ops', true) }}"
     db2wh_instance_name: "{{ lookup('env', 'DB2WH_INSTANCE_NAME') }}"
-    db2wh_load_from_config: "{{ lookup('env', 'DB2WH_db2wh_load_from_config') | default('false', true) }}"
+    db2wh_load_from_config: "{{ lookup('env', 'DB2WH_LOAD_FROM_CONFIG') | default('false', true) }}"
```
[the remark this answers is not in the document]
andrercm: fixed.

pipelines/tasks/mas/hack-manage-db2.yaml
andrercm: @alequint would you be able to fix this part when you refactor the pipelines? Don't want to touch in things I know you're working on.

andrercm: Just removed it @durera

durera:
Going to proceed with the merge of this into master, final tuning ahead of v5.1 release will take place in other PRs if needed, but this one is already far too big and unwieldly.

durera:
Just waiting on the quay.io outage to be resolved before merging this to master ... https://status.quay.io/

durera (approved):

PR #106, opened 2021-12-16
[patch] Add BAS storage classes validation

andrercm:
The BAS install role does not check for required parameters being provided, when these are missed instead of providing the user with a nice clear error message the BAS install will fail and they must debug what happened.

This update will ensure that the installation is aborted instantly if either of the required storage classes are not defined.

### DIFF
--- a/ibm/mas_devops/playbooks/dependencies/install-bas.yml
+++ b/ibm/mas_devops/playbooks/dependencies/install-bas.yml
@@ -5,8 +5,8 @@
     # BAS Configuration
     bas_namespace: "{{ lookup('env', 'BAS_NAMESPACE') | default('ibm-bas', true) }}"
 
-    bas_persistent_storage_class: "{{ lookup('env', 'BAS_PERSISTENT_STORAGE_CLASS') | default('ibmc-file-bronze-gid') }}"
-    bas_meta_storage_class: "{{ lookup('env', 'BAS_META_STORAGE_CLASS') | default('ibmc-block-bronze')}}"
+    bas_persistent_storage_class: "{{ lookup('env', 'BAS_PERSISTENT_STORAGE_CLASS') }}"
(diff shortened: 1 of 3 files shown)


ibm/mas_devops/playbooks/dependencies/install-bas.yml
```diff
@@ -5,8 +5,8 @@
     # BAS Configuration
     bas_namespace: "{{ lookup('env', 'BAS_NAMESPACE') | default('ibm-bas', true) }}"
 
-    bas_persistent_storage_class: "{{ lookup('env', 'BAS_PERSISTENT_STORAGE_CLASS') }}"
-    bas_meta_storage_class: "{{ lookup('env', 'BAS_META_STORAGE_CLASS') }}"
+    bas_persistent_storage_class: "{{ lookup('env', 'BAS_PERSISTENT_STORAGE_CLASS') | default('ibmc-file-bronze-gid') }}"
```
durera, line 8: You can't do this .. as I mentioned before .. this playbook is not specific to ROKS, these classes only exist on ROKS.  

- If a user calls this playbook directly they have to choose a storage class.  
- If a user calls this playbook via the various `-roks` playbooks then the class will have been set.

Unless the playbook has `-roks` in it's name there can be no assumptions about ROKS storages classes being available.

durera (changes requested):
Other than the defaulting of ROKS storage classes in a generic playbook, the other changes are nice and bring this into line with other roles that do safety checks on required parameter 👍

ibm/mas_devops/playbooks/dependencies/install-bas.yml
andrercm: fixed, removed default storage classes for ROKS only. made a small correction to the docs as well as it's not going to be used per default anymore.

PR #114, opened 2021-12-23
Make sure hack-manage playbook is looking at the right namespace

alequint:
- Fix how hack-manage playbook look for db2 pod considering different namespaces (cpd35 x cpd40)
- Fix configure-ocp on checking config map for ocp version


ibm/mas_devops/roles/cp4d_install_services/tasks/cpd40/cpd-install-services.yml
```diff
@@ -24,6 +24,12 @@
     msg: "cpd_block_storage_class property has not been set but it is required to install IBM Watson Discovery"
   with_items: "{{ cpd_services }}"
 
+# Note: this role is always called together with cpd_install and it is because of this that
```
durera, line 27: That's not true, at least to best of my understanding.  

The reason they are seperate is so that a user can independently call `cpd_install_services` and `cpd_install` roles ... e.g. today you could `cpd_install` and `cpd_install_services` (db2) .. then tomorrow you could `cpd_install_services` (watson studio) ... we can't rely on variables that are only set in another role being inherited. 

However, lookign at the code I think the interpretation of how it works is wrong in general.  The reason it works is that all of our playbooks that call this role explicitly define the `cpd_meta_namespace` variable ... it's not due to any link to the cpd_install role.  

tldr; I think this comment block is wrong, but I could be mistaken :)

ibm/mas_devops/roles/cp4d_db2wh_manage_hack/tasks/main.yml
```diff
@@ -2,13 +2,39 @@
 # Should should be part of the manage operator, but is not so we have to do it
 # as a seperate step in the install flow for now :(
 
+# 0.1. Check which CPD version has been installed (in case cpd_version is not informed)
+# -----------------------------------------------------------------------------
+- name: Check what CPD version has been installed by looking at the namespaces
+  community.kubernetes.k8s_info:
+    api_version: v1
+    kind: Namespace
+    name: cpd-services
```
durera, line 11: in cpd v3.5 the namespace cpd was installed to was configurable .. have we removed the ability to configure where it's installed with cpd v4, as this logic will only work if the namespace **has** to be cpd-services, and there is no possibility to configure it otherwise.

ibm/mas_devops/roles/ocp_setup_mas_deps/tasks/main.yml
```diff
@@ -18,7 +18,7 @@
     - openshift_config.resources is defined
     - openshift_config.resources | length > 0
   set_fact:
-    ocp_install_invoker: "{{ openshift_config.resources[0].data.invoker }}" # Known values: "ROKS"
+    ocp_install_invoker: "{{ openshift_config.resources[0].data.invoker }}" # Known values: "ROKS" or "user" (quickburn)
```
durera, line 21: The addition to the comment is incorrect (unless we have new evidence that sometimes Fyre uses `openshift-install` instead of `openshift-install-manifest`) ... in this scenario only ROKS is possible, From our testing Fryre doesn't use this configmap, it uses the `openshift-install-manifests` configmap that we check if this one doesn't exist.  see snippet below where we comment on the known possible values in the alternative configmap.  However it may be mroe complicate than that .. e.g. a specific version of OCP on Fyre uses different configmaps perhaps?  What version of OCP on Fyre did we find this configmap being used on that triggered this update to the doc?

```
# If we didn't already set the new facts then it means we couldn't find the "openshift-install" configmap
- name: Get OCP Install Information (openshift-install-manifests)
  when: ocp_install_invoker is not defined
  community.kubernetes.k8s_info:
    api_version: v1
    kind: ConfigMap
    name: openshift-install-manifests
    namespace: openshift-config
  register: openshift_config
- name: Register OCP install configuration variables (openshift-install-manifests)
  when:
    - ocp_install_invoker is not defined
    - openshift_config.resources is defined
    - openshift_config.resources | length > 0
  set_fact:
    ocp_install_invoker: "{{ openshift_config.resources[0].data.invoker }}" # Known values: "user" (in Fyre)
    ocp_install_version: "{{ openshift_config.resources[0].data.version }}"
ocp_install_release: "{{ openshift_config.resources[0].data.version | regex_search('[^.]*.[^.]*') }}"
```

durera (changes requested):
Also, we need to get people updating the docs as part of these kinds of updates - it can't be all on me - can you please update the doc for the role you are fixing to bring it up to scratch:

- https://ibm-mas.github.io/ansible-devops/playbooks/mas/#manage-db2-hack
- https://ibm-mas.github.io/ansible-devops/roles/cp4d_db2wh_manage_hack/

See an example of the expected template for the playbook and role doc here:

- https://ibm-mas.github.io/ansible-devops/playbooks/dependencies/#install-amq-streams
- https://ibm-mas.github.io/ansible-devops/roles/amqstreams/

Once the doc is updated, we can update #3 to mark another topic off the list.

ibm/mas_devops/roles/cp4d_db2wh_manage_hack/tasks/main.yml
alequint: In cpd4.0 I noticed the vars are loaded and overwrite whatever is in cpd namespace environment var, different from cpd 3.5 (where it uses whatever was sent by users)... my initial thought was to remove the namespace to be configurable for 3.5, but then thouht someone may be relying on it already so I kept it as it is, but did not include the option to do the same for 4.0, keeping the code as it was originally created, loading internal vars, etc

But it does not mean we can introduce the configurable namespaces for 4.0 in future patches if needed (then we will need two envvars, one for cpd, another one for services, etc), not necessarily part of this specific PR though. Makes sense?

ibm/mas_devops/roles/ocp_setup_mas_deps/tasks/main.yml
alequint: You are right @durera ... I got the config map from Jean (who had a fyre cluster), but did not notice the name of the config-map was different (checking again now, and I confirmed it was -manifest one). And now it makes sense your change to include this new config map checking. I will remove this piece of comment!

ibm/mas_devops/roles/cp4d_install_services/tasks/cpd40/cpd-install-services.yml
alequint: > The reason it works is that all of our playbooks that call this role explicitly define the cpd_meta_namespace variable .

It's not I see today... here lite manage for example: https://github.com/ibm-mas/ansible-devops/blob/master/ibm/mas_devops/playbooks/lite-manage-roks.yml#L46-L49 - and even if we did, for cpd 4.0 cpd_meta_namespace would not be enough, as we have two namespaces

> we can't rely on variables that are only set in another role being inherited; if that is the case it's a major bug.

The cpd 4.0 vars were only loaded here: https://github.com/ibm-mas/ansible-devops/blob/master/ibm/mas_devops/roles/cp4d_install/tasks/cpd40/cpd-install.yml#L5 and there is nothing else that pass such values to the other role (called, usually, right after this one by our cpd playbooks)

So the comment is more to raise such TODO to fix, as I took some time to figure it out, instead of fixing that now and then need to retest everything again - we can do that in a subsequent patch (and then more tests), so we do not hold this stuff for so long. Makes sense?

ibm/mas_devops/roles/ocp_setup_mas_deps/tasks/main.yml
alequint: Fixed

ibm/mas_devops/roles/cp4d_install_services/tasks/cpd40/cpd-install-services.yml
durera: > It's not I see today... here lite manage for example: https://github.com/ibm-mas/ansible-devops/blob/master/ibm/mas_devops/playbooks/lite-manage-roks.yml#L46-L49

The way that works is through variable priority:

1. lite-manage-roks imports the install-services-db2 playbook and doesn't set any value for `cpd_meta_namespace` in it's vars, if it did this would override anything set in the playbook we import, or in the defaults built into the role
2. the imported install-services-db2 playbook sets a value for `cpd_meta_namespace` (based on `CPD_NAMESPACE` environment variable, defaulting to `cpd-meta-ops` if that env var is not set), so that's what we use, which would override any defaults built into the role (if we had one), but would itself be overridden if we were to set a value in the lite-manage-roks playbook import.

Because there is no setting in defaults/main.yml we are effectively declaring `cpd_meta_namespace` as a required variable that must be passed in whenever running this role.  There's no connection to the cpd-install role other than that they both happen to be called in the same playbook in our sample playbooks.

ibm/mas_devops/roles/cp4d_db2wh_manage_hack/tasks/main.yml
durera: Okay, given this info I think the most sensible thing to do is to remove that variable for 3.5 as well and just have it so it always installs to a set namespace, so we can offer a consistent behaviour across the two versions.  

I see minimal value in customising the install directory of cp4d, we've never used it ourselves; and keeping it so that the env var works sometimes and not others depending on what version of cp4d is being installed is not good given part of the job of the collection is to allow users to not have to think about what version of cp4d to install as we will work it out from the mas_channel etc.

ibm/mas_devops/roles/cp4d_install_services/tasks/cpd40/cpd-install-services.yml
alequint: 1 and 2 above are correct, yes, but let's continue the flow:

1. lite-manage-roks imports install-services-db2 and do not set the cpd_meta_namespace var
2. install-services-db2 defaults cpd_meta_namespace to either the environment var or to cpd-meta-ops if not set, ok.
3. cpd-install role starts. If cpd version is 35, we use the cpd_meta_namespace defaulted in the playbook, no problem
3.1 if cpd version is 4, we load vars [here](https://github.com/ibm-mas/ansible-devops/blob/master/ibm/mas_devops/roles/cp4d_install/tasks/cpd40/cpd-install.yml#L5), and now cpd_meta_namespace is overwritten AND a new var appears: `cpd_services_namespace` - in other words we are ignoring what have been passed, but let's move forward
4. cpd-isntall-services role starts... if cpd 35, we will continue using the var defaulted in playbook. But, if it is 4.0, we are using the vars from 3.1... otherwise, `cpd_services_namespace` would not even exist as nobody set that besides the cpd-install

PR #123, opened 2021-12-28
[patch] Increase manage appws timeout and fix db2wh_svc_fqn generation

durera:
A couple of small fixes and updates to developer documentation:
- Add 1 hour to manage workspace config timeout.  Manage application workspace configuration takes a really long time, increase the timeout from 3 to 4 hours.
- Fix generation of `db2wh_svc_fqn` variable, this was still referring to a configmap lookup instead of simply computing it from the one value that we need to lookup - the instance ID.


alequint:
@durera sometimes even more (5 or 6 hours), depending on how many industry solutions will be installed.

Issue #124, opened 2021-12-29
Utilize ansible assert instead of fail / when


durera:
Refer to: https://docs.ansible.com/ansible/latest/collections/ansible/builtin/assert_module.html

Our roles would be better implemented like so:

```yaml
# 1. Check for required parameters
# -----------------------------------------------------------------------------
- name: "Assert that xxx is provided"
  assert:
    that:
      - xxx is defined
      - xxx != ""
    fail_msg: "xxx property is required"
```

instead of using the fail task guarded by a when.  This will result in clearer results as we won't have numerous skipped steps on success, instead the task will actually be marked as a success/failure.

When taking on this issue directly we should make this change globally; in addition all new code should follow this convention, and all updates to existing code should apply these changes as part of their work as a condition of acceptance.

PR #133, opened 2022-01-03
[patch] db2wh backup & restore

andrercm:


ibm/mas_devops/playbooks/cp4d/db2wh-backup-restore.yml
```diff
@@ -0,0 +1,13 @@
+---
```
durera, line 1: Can you move the vars / defaults into the role (e.g. see example here https://github.com/ibm-mas/ansible-devops/blob/master/ibm/mas_devops/roles/cp4d_db2wh/defaults/main.yml)

docs/roles/cp4d_db2wh_backup.md
```diff
@@ -0,0 +1,63 @@
...
+At the end of the backup process, you will find the required files to run a successful restore process in the chosen `DB2WH_BACKUP_DIR`:
+
+- DB2 backup files i.e `BLUDB.0.db2inst1.DBPART000.202XXXXXXXXXXX.001`
+- DB2 keystore files (.p12 and .sth)
+- DB2 instance master key label file (.kdb)
+
+Role Variables
+--------------
+
+### Required Variables
```
durera, line 14: Can you update the README to match the new(ish) format we are using now: e.g. https://github.com/ibm-mas/ansible-devops/blob/master/ibm/mas_devops/roles/cp4d_db2wh/README.md#role-variables

docs/roles/cp4d_db2wh_restore.md
```diff
@@ -0,0 +1,65 @@
...
+This role runs a restore procedure onto a CloudPak for Data DB2 Warehouse instance.
+In order to begin the restore process, you must have the required files to run a successful restore process in the chosen `DB2WH_BACKUP_DIR`:
+
+- DB2 backup files i.e `BLUDB.0.db2inst1.DBPART000.202XXXXXXXXXXX.001`
+- DB2 keystore files (.p12 and .sth)
+- DB2 instance master key label file (.kdb)
+
+Note: These files are generated automatically if you run `ibm.mas_devops.cp4d_db2wh_backup` role. If any of the above files are not found in the `DB2WH_BACKUP_DIR`, the restore process will fail.
+
+Role Variables
```
durera, line 13: Can you update the README to match the new(ish) format we are using now: e.g. https://github.com/ibm-mas/ansible-devops/blob/master/ibm/mas_devops/roles/cp4d_db2wh/README.md#role-variables

ibm/mas_devops/playbooks/cp4d/db2wh-backup-restore.yml
```diff
@@ -0,0 +1,13 @@
...
+- hosts: localhost
```
durera, line 2: It would be better to seperate this into two playbooks ... backup and restore as seperate things ... then in pipelines etc you can either backup only, restore only, or backup followed by a restore.

ibm/mas_devops/playbooks/cp4d/db2wh-backup-restore.yml
```diff
@@ -0,0 +1,13 @@
...
+  any_errors_fatal: true
+  vars:
```
durera, line 4: All these should end up living in the role's default/main.yml ... see earlier examples provided

ibm/mas_devops/roles/cp4d_db2wh_backup/tasks/main.yml
```diff
@@ -0,0 +1,126 @@
...
+- name: "DB2 Backup - Debug information"
+  debug:
+    msg:
+      - "CPD Namespace .......................... {{ cpd_meta_namespace_source }}"
+      - "DB2 Instance ID ........................ {{ db2wh_instance_id_source }}"
+      - "DB2 Backup Files Directory ............. {{ db2wh_backup_folder }}"
+
+# 2. Fail if no required keys are provided
+# -----------------------------------------------------------------------------
+- name: "Fail if cpd_meta_namespace_source has not been provided"
```
durera, line 13: Could you use the new/better convention using assert for these ... see issue https://github.com/ibm-mas/ansible-devops/issues/124

Example of them in use: https://github.com/ibm-mas/ansible-devops/blob/upgrade87/ibm/mas_devops/roles/suite_upgrade/tasks/main.yml#L5

ibm/mas_devops/roles/cp4d_db2wh_backup/tasks/main.yml
```diff
@@ -0,0 +1,126 @@
...
+
+- name: "Debug information"
+  debug:
+    msg:
+      - "Pod name ........... {{ db2wh_pod_name }}"
+  failed_when: db2wh_pod_name is not defined or db2wh_pod_name == ""
+
+# 4. Run backup script in source db2wh pod
+# -----------------------------------------------------------------------------
+- name: "Running backup script on DB2WH pod {{ db2wh_instance_id_source }} as db2inst username... (this can take several minutes)"
```
durera, line 51: Can we break this up into multiple steps, it will be much easier to debug if each `oc exec` is it's own step instead of all munged together ... e.g. see the changes I made in the db2 hack script where we now have a step to copy the file into the container and a step to execute it: https://github.com/ibm-mas/ansible-devops/blob/master/ibm/mas_devops/roles/cp4d_db2wh_manage_hack/tasks/main.yml#L120-L128

We should also use the oc copy command and a template/local file to get the file into the container (also see the example I provided in the link above) rather than inlining the file in the task itself, makes things easier to read for the user if they for example want to use the role as a guide rather than execute it directly.

ibm/mas_devops/roles/cp4d_db2wh_restore/tasks/main.yml
```diff
@@ -0,0 +1,241 @@
...
+  debug:
+    msg:
+      - "CPD Namespace .......................... {{ cpd_meta_namespace_target }}"
+      - "DB2 Instance ID ........................ {{ db2wh_instance_id_target }}"
+      - "DB2 Backup Files Directory ............. {{ db2wh_backup_folder }}"
+      - "DB2 Backup Temp Log Files Directory .... /tmp/db_backup_logs"
+
+# 2. Fail if no required keys are provided
+# -----------------------------------------------------------------------------
+- name: "Fail if cpd_meta_namespace_target has not been provided"
```
durera, line 14: see earlier comments about using assert module instead

durera (changes requested):

ibm/mas_devops/playbooks/cp4d/db2wh-backup-restore.yml
andrercm: Fixed.

docs/roles/cp4d_db2wh_backup.md
andrercm: Fixed.

docs/roles/cp4d_db2wh_restore.md
andrercm: Fixed.

ibm/mas_devops/playbooks/cp4d/db2wh-backup-restore.yml
andrercm: Fixed.

ibm/mas_devops/playbooks/cp4d/db2wh-backup-restore.yml
andrercm: Fixed.

ibm/mas_devops/roles/cp4d_db2wh_backup/tasks/main.yml
andrercm: Fixed.

ibm/mas_devops/roles/cp4d_db2wh_backup/tasks/main.yml
andrercm: fixed.

ibm/mas_devops/roles/cp4d_db2wh_restore/tasks/main.yml
andrercm: Fixed.

ibm/mas_devops/roles/cp4d_db2wh_backup/defaults/main.yml
```diff
@@ -0,0 +1,4 @@
+# db2 backup default vars
+db2wh_backup_dir: "{{ lookup('env', 'DB2WH_BACKUP_DIR') }}"
+cpd_meta_namespace_source: "{{ lookup('env', 'CPD_NAMESPACE_SOURCE') }}"
```
durera, line 3: Will this work with cpd v3 and v4 .. in v4 the db2 pod is not in the meta namespace, but a seperate services namespace?  e.g. could we do with some code like this: https://github.com/ibm-mas/ansible-devops/blob/master/ibm/mas_devops/roles/cp4d_db2wh/tasks/main.yml#L7-L45 somewhere in this role?

ibm/mas_devops/roles/cp4d_db2wh_backup/tasks/main.yml
```diff
@@ -0,0 +1,133 @@
...
+# 2. Fail if no required keys are provided
+# -----------------------------------------------------------------------------
+- name: "Assert that cpd_meta_namespace_source is defined"
+  assert:
+    that:
+      - cpd_meta_namespace_source is defined
+      - cpd_meta_namespace_source != ""
+    fail_msg: "cpd_meta_namespace_source property is required"
+
+- name: "Assert that db2wh_instance_id_source is defined"
```
durera, line 20: Is this definitely an instance id versus instance name?  We use both vars in other db2 roles and just want to check we are being consistent with our naming conventions here :) 

In cpd4 instance id == instance name and we determine the id from the name input (https://github.com/ibm-mas/ansible-devops/blob/master/ibm/mas_devops/roles/cp4d_db2wh/tasks/cpd40/provision-instance.yml#L171), but in cpd3 they are different because the API is being used to create the database and again we use the name to determine the instance id, but rely on the additional configmap created during deploy.

if this needs to be instance id, I think we should have the same logic we have in cp4d_db2wh in here to convert from db2wh_instance_name to the ID that we need

ibm/mas_devops/roles/cp4d_db2wh_restore/README.md
```diff
@@ -0,0 +1,57 @@
...
+Role Variables
+--------------
+
+### db2wh_backup_dir
+Required. Local directory that stores the backup files to be used in the DB2 Warehouse restore process, i.e `/Users/Documents/db_backup`.
+
+- Environment Variable: `DB2WH_BACKUP_DIR`
+- Default: None
+
+### cpd_meta_namespace_target
```
durera, line 22: See earlier comments, I think we can remove this entirely and use the smart auto-detect namespace code that lives in cp4d_db2wh in this role so that the user doesn't have to worry about the cpd namespace input at all.

ibm/mas_devops/roles/cp4d_db2wh_restore/README.md
```diff
@@ -0,0 +1,57 @@
...
+Required. CloudPak for Data namespace for the target DB2 Warehouse instance where you want to restore the backup to, i.e `cpd-services`.
+
+- Environment Variable: `CPD_NAMESPACE_TARGET`
+- Default: None
+
+### db2wh_instance_id_target
```
durera, line 28: See earlier comments, is there a way we can make this consistent with the inputs for the main cp4d_db2wh role and have `db2wh_instance_name` be the input, and we can determine the id from the name as we do in e.g. hack_manage script.

ibm/mas_devops/roles/cp4d_db2wh_restore/README.md
```diff
@@ -0,0 +1,57 @@
...
+Required. DB2 Warehouse target instance to restore the backup to, i.e `db2wh-1641225392061935`.
+
+- Environment Variable: `DB2WH_INSTANCE_ID_TARGET`
+- Default: None
+
+### oc_login_target
```
durera, line 34: Should this be needed?  the scope of this role is restore to the current target cluster (as all roles are designed to target the current cluster).  In a workflow/pipeline where the user wants to backup in one cluster, rtestore to another, they would chain roles together like:
- ocp_login (cluster1)
- db2wh_backup
- ocp_login (cluster2)
- db2wh_backup

I think we can safely remove this to simplify this role, and rely on the chaining of roles to support this scenario.

ibm/mas_devops/roles/cp4d_db2wh_restore/tasks/main.yml
```diff
@@ -0,0 +1,286 @@
+---
+# 1. Provide debug information to the user
+# -----------------------------------------------------------------------------
+- name: "DB2 Restore - Debug information"
+  debug:
+    msg:
+      - "CPD Namespace .......................... {{ cpd_meta_namespace_target }}"
+      - "DB2 Instance ID ........................ {{ db2wh_instance_id_target }}"
+      - "DB2 Backup Files Directory ............. {{ db2wh_backup_dir }}"
+      - "DB2 Backup Temp Log Files Directory .... /tmp/db_backup_logs"
```
durera, line 10: This string appears in multiple places in the role, I think it would make sense to move this to a variable in defaults, to make it easy to update should this need to be changed in the future so it's only in one place.

durera (changes requested):
Review 2

ibm/mas_devops/roles/cp4d_db2wh_backup/defaults/main.yml
andrercm: There's no backup/restore process change whether this is a v3.5 or v4.0, it all behaves the same no matter cpd version, during tests I even used a source db2wh from v3.5 and restored in a target db2wh v4.0 to ensure, and it works fine either way.

ibm/mas_devops/roles/cp4d_db2wh_backup/tasks/main.yml
andrercm: Here we just need something to lookup for the `db2wh` main pod, so it would be something as `db2wh-1637258370280916` (in case of db2wh created via api, cpd v3.5) or `db2wh-iot` (in case of db2wh created via db2ucluster, cpd v4)
Regardless for the case, do we need to add this conversion here? I'm thinking if we need to use this playbook in conjuction with another, can't we pass the `db2wh_instance_id_source` with the expected converted output from previous defined by other playbooks?
Just not sure if worths adding such cpd_version conversion for a role/playbook that works the same way regardless of cpd_version.

ibm/mas_devops/roles/cp4d_db2wh_backup/defaults/main.yml
andrercm: fixed to be same as in `cp4d_db2wh` and `cp4d_db2wh_manage_hack` as you suggested.

ibm/mas_devops/roles/cp4d_db2wh_backup/tasks/main.yml
andrercm: fixed to be same as in `cp4d_db2wh` and `cp4d_db2wh_manage_hack` as you suggested.

ibm/mas_devops/roles/cp4d_db2wh_restore/README.md
andrercm: fixed to be same as in `cp4d_db2wh` and `cp4d_db2wh_manage_hack` as you suggested.

ibm/mas_devops/roles/cp4d_db2wh_restore/README.md
andrercm: fixed to be same as in `cp4d_db2wh` and `cp4d_db2wh_manage_hack` as you suggested.

ibm/mas_devops/roles/cp4d_db2wh_restore/README.md
andrercm: yup, removed the need of it, haven't considered that we can leverage `ocp_login` role in between, good call.

ibm/mas_devops/roles/cp4d_db2wh_restore/tasks/main.yml
andrercm: fixed it, all multiple string occurences for path inside db2u pod have been moved to default vars.

ibm/mas_devops/roles/cp4d_db2wh_restore/tasks/main.yml
```diff
@@ -0,0 +1,320 @@
...
+  ansible.builtin.template:
+    src: postRestore.sh.j2
+    dest: /tmp/postRestore.sh
+    mode: '777'
+
+- name: Copy the postRestore.sh script into the db2 pod {{ db2wh_pod_name }}
+  shell: "oc cp /tmp/postRestore.sh {{ db2wh_pod_namespace }}/{{ db2wh_pod_name }}:/tmp/postRestore.sh"
+
+- name: Run postRestore.sh script on db2 pod {{ db2wh_pod_name }}
+  shell: oc exec -n {{ cpd_meta_namespace_target }} {{ db2wh_pod_name }} -- su -lc '/tmp/postRestore.sh | tee /tmp/postRestore.log' db2inst1
```
ctrafa, line 309: cpd_meta_namespace_target must be replaced by db2wh_pod_namespace

ibm/mas_devops/roles/cp4d_db2wh_restore/tasks/main.yml
```diff
@@ -0,0 +1,320 @@
...
+  failed_when: post_restore_output.rc != 0
+
+- debug:
+    msg: "{{ post_restore_output.stdout_lines }}"
+
+# 13. Delete temporary db2 backup files in target db2wh pod
+# -----------------------------------------------------------------------------
+- name: "Delete db_backup folder from {{ db2wh_pod_name }} now that backup files were restored..."
+  shell: |
+    oc exec -it -n {{ cpd_meta_namespace_target }} {{ db2wh_pod_name }} -- bash -c "sudo rm -rf {{ db2wh_temp_backup_dir }} && sudo rm -rf {{ db2wh_temp_backup_logs_dir }} && sudo rm -rf {{ db2wh_db2inst1_backup_dir }}"
```
ctrafa, line 320: cpd_meta_namespace_target must be replaced by db2wh_pod_namespace

ctrafa (changes requested):
cpd_meta_namespace_target must be replaced by db2wh_pod_namespace at lines 309 and 320 of ansible-devops/ibm/mas_devops/roles/cp4d_db2wh_restore/tasks/main.yml

ibm/mas_devops/roles/cp4d_db2wh_restore/tasks/main.yml
andrercm: @ctrafa this is fixed

ibm/mas_devops/roles/cp4d_db2wh_restore/tasks/main.yml
andrercm: @ctrafa this is fixed

ctrafa (approved):

andrercm:
@durera all reviews are completed now. Seems ready to be merged.

andrercm:
@alequint can you help merge it? all PR reviews are through now.

PR #135, opened 2022-01-05
BAS Scheduler Event + Docs

alequint:
- Moved vars from BAS playbook to BAS role
- Added a new BAS attribute `bas_event_scheduler_frequency`
- Review role and playbook docs to follow new format


alequint:
Bad branch. Closing.

PR #157, opened 2022-02-03
[minor] Miscellaneous cp4d_db2wh role updates

cuddlyporcupine:
- enable user to configure DB2_WORKLOAD
- enable user to deploy db2wh instance on a dedicated worker node (taint/label node) and deploy with node affinity
- fix bug in default yaml for db2wh_dbname variable
- route object using reencrypt gives DRDA db2 comms error when attempting to connect, changing route type to passthrough resolves error


durera (approved):
LGTM

PR #161, opened 2022-02-09
[patch] Fix SLS Bootstrap verification

### LINKED ISSUE #130: SLS role does not verify SLS bootstrap success
When using `ibm.mas_devops.sls` to bootstrap a license ID and or key the role should assert that the bootstrap was successful.  Currently it does not, allowing the role to report success when the bootstrap failed.

e.g.  after running the role the LicenseService was left in this state, but the role reported back success.  This caused the deployment pipeline to fail much later than it should have.

````
oc -n ibm-sls get licenseservice
NAME   VERSION   STATUS   INITIALIZED            LICENSEID      REGISTRATIONKEY               AGE
sls    3.2.4     Ready    MissingConfiguration   <redacted>     <redacted>                    8h
```

In this scenario (easily reproduced by using any invalid licence file content) the role should fail because the initialized status condition is indicating that the bootstrap did not succeed.

caroazad:
- Fixes #130


durera (changes requested):
Validation should use Assert rather than fail/when ... this leads to a bbetter user experience/clearer reporting of success/failure than our previous approach ... see #124 for more dfetails.

Additionally, the Initilaized state should only be checked if the user chose to bootstrap, which they may not have done - in which case initialized = false is acceptable, we should have something like:

```
name: Verify bootstrap success
assert:
  when: bootstrap is defined
  that: <initialized == true>
  failure_message: Bootstrap failed
```

The overall logic in this verify needs to be examined, as it already looks like it's trying to check all these states, it's just not working ... lines 38-49 here are already (trying to) check that initialized is set to the expected value, so I don't think we even need to add new checks, just work out what's wrong with the existing ones and fix them.

caroazad:
> Validation should use Assert rather than fail/when ... this leads to a bbetter user experience/clearer reporting of success/failure than our previous approach ... see #124 for more dfetails.
> 
> Additionally, the Initilaized state should only be checked if the user chose to bootstrap, which they may not have done - in which case initialized = false is acceptable, we should have something like:
> 
> ```
> name: Verify bootstrap success
> assert:
>   when: bootstrap is defined
>   that: <initialized == true>
>   failure_message: Bootstrap failed
> ```
> 
> The overall logic in this verify needs to be examined, as it already looks like it's trying to check all these states, it's just not working ... lines 38-49 here are already (trying to) check that initialized is set to the expected value, so I don't think we even need to add new checks, just work out what's wrong with the existing ones and fix them.

Okay, I will work on that. Thank you for the feedback

caroazad:
From your feedback, I looked at the existing code for the initialization verification and included the test for if the initialized status value was not only available but was equal to True. I don't know if this is enough but I tested it on both positive and negative situations.

PR #168, opened 2022-02-18
[patch] fix cpd v4 storage class and mas channel

andrercm:
This PR addresses two things:
- Fixes storage classes for CPD v4 to not use block storage class anymore. Recently we discovered that CPD v4 docs are a bit controversy. Even though in their official [docs](https://www.ibm.com/docs/en/cloud-paks/cp-data/4.0?topic=installing-cloud-pak-data ) they mention about the use of block storage in their sample CRs, the cpd support team told us this is not how cpd team test it or recommend as supported. In a recent conversion, they pointed us to this [docs](https://www.ibm.com/docs/en/cloud-paks/cp-data/4.0?topic=requirements-storage) which describes the storage classes available for ROKS which DOES NOT include `ibmc-block-gold` as supported storage class. I tested the installation with `ibmc-file-gold-gid` as suggested by them and it works fine, so I'm adjusting our collection to use the recommended.
- With MAS 8.7 being release, we need to change the way our smart logic works to auto determine which cpd version to install based on the `mas_channel` being used along with the playbooks. In summary I'm changing the logic a bit to only install cpd v3.5 if `mas_channel=8.6.x`, anything other than that, we will install cpd v4.0.


ibm/mas_devops/roles/cp4d_install/templates/cpd40/ibmcpd.yaml
```diff
@@ -7,5 +7,7 @@ spec:
   license:
     accept: true
     license: Standard
+  # both storage classes for cpd should be ibmc-file-gold-gid
+  # https://www.ibm.com/docs/en/cloud-paks/cp-data/4.0?topic=requirements-storage
   storageClass: "{{ cpd_storage_class }}"
-  zenCoreMetadbStorageClass: "{{ cpd_block_storage_class }}"
+  zenCoreMetadbStorageClass: "{{ cpd_storage_class }}"
```
mikehollinger, line 13: i would suggest tbh doing something different here; instead of using the same storage class from elsewhere, name the variable to match the intent, which is to set the metastore storage class, and set the var explicitly elsewhere to some value. It might happen that we'd use the same in multiple spots, but in the future, if there's ever a change to a different class which has an odd side effect, we can make changes in one spot, vs having to fiddle and search around in the depths of the roles?

andrercm: @mikehollinger maybe what we can do then is to set `"cpd_block_storage_class: "{{ cpd_storage_class }}"` [here](https://github.com/ibm-mas/ansible-devops/blob/master/ibm/mas_devops/roles/cp4d_install/defaults/main.yml#L14), making a comment about this and removing block storage usage from our ansible docs for now. That way, if we ever need to change it, it would be in this single place. 
The reason why I think it's not a good idea to set explicitly a value for it is that it would only work for ROKS, but we could use this playbook for other types of envs like Fyre/Quickburn and even AWS/ROSA in the future, so we need to leave it flexible as we can. What u think?

Issue #162, opened 2022-02-10
Publishing a tag does to create a release of the docker image


durera:
When master branch builds we create `quay.io/ibmmas/ansible-devops:x.y.z-pre.master`, however currently there is nothing in the tag build that will create and publish the `quay.io/ibmmas/ansible-devops:x.y.z` image.

Workaround for now is for someone with push permissions to do this:

```
docker pull quay.io/ibmmas/ansible-devops:x.y.z-pre.master
docker tag quay.io/ibmmas/ansible-devops:x.y.z-pre.master quay.io/ibmmas/ansible-devops:x.y.z
docker push quay.io/ibmmas/ansible-devops:x.y.z
```

This is easily fixed in the github actions defintions when we have a chance (byt building and publishing the container image).

PR #177, opened 2022-03-08
[patch] Fix until condition on kafka resource

andrercm:
- Fix for timeout condition when waiting for Kafka cluster to be provisioned
- Implements #162

### DIFF
--- a/ibm/mas_devops/roles/amqstreams/tasks/main.yaml
+++ b/ibm/mas_devops/roles/amqstreams/tasks/main.yaml
@@ -69,25 +69,21 @@
 # We have seen instances where the Kafka cluster is reported as Ready but
 # the necessary fields in the status resource are not set yet so we retry until
 # the expected fields are set as well
-- name: "Lookup Kafka Resource"
+- name: "Wait for Kafka Resource to be ready (60s delay)"
   community.kubernetes.k8s_info:
     api_version: kafka.strimzi.io/v1beta2
     kind: Kafka
     name: "{{ kafka_cluster_name }}"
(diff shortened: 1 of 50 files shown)


ibm/mas_devops/roles/amqstreams/tasks/main.yaml
```diff
@@ -69,24 +69,21 @@
...
   register: kafka_cluster_lookup
-  retries: 120 # Once the cluster is ready we will wait for up to 20 minutes for the listener information to be added to the status sub-resource
-  delay: 10 # seconds
+  retries: 30 # will wait for up to 30 minutes for the listener information to be added to the status sub-resource
+  delay: 60 # seconds
   until:
+    - kafka_cluster_lookup.resources is defined
     - kafka_cluster_lookup.resources | length > 0
     - kafka_cluster_lookup.resources[0].status is defined
+    - kafka_cluster_lookup.resources[0].status.conditions[0].type == "Ready"
```
durera, line 85: This seems risky ... the first condition might not always be the ready condition.  We should use code something like this to check for presence of a Ready condition anywhere in the status, and then confirm the status is set to "True":

```
- kafka_cluster_lookup.resources[0].status.conditions | selectattr('type', 'equalto', 'Ready') | map(attribute='status') | list | length > 0
- kafka_cluster_lookup.resources[0].status.conditions | selectattr('type', 'equalto', 'Ready')| map(attribute='status') | list | first == "True"
```

durera (changes requested):

durera (approved):

PR #180, opened 2022-03-09
[patch] Add multicloud-bootstrap for AWS

amitmangalvedkar:
Initially, MultiCloud bootstrap focuses on the framework required to leverage this automation project for use in AWS.

The scripts are not meant to be called manually unless needed for troubleshooting, they are called in a specific order during the bootstrap process. The bootstrap process is called from the virtual server (aka the `bootnode`) automatically when the bootnode is created.


amitmangalvedkar:
@durera Can you please review this code

durera (changes requested):
Please can you add a README.md into the multicloud-bootstrap directory that describes the content, what it does, how to use it etc.  A good example is this one here: https://github.com/ibm-mas/ansible-devops/blob/master/pipelines/README.md ... we should have something like this for the multicloud-bootstrap content

multicloud-bootstrap/init.sh
```diff
@@ -0,0 +1,296 @@
...
+  log "ERROR: Incorrect value for OFFERING_TYPE - $OFFERING_TYPE"
+  PRE_VALIDATION=fail
+fi
+
+## Variables
+# OCP variables
+export GIT_REPO_HOME=$(pwd)
+export CLUSTER_NAME="masocp-${RANDOM_STR}"
+export OPENSHIFT_USER="masocpuser"
+export OPENSHIFT_PASSWORD=masocp${RANDOM_STR}pass
```
Ashish-Mahawal1, line 68: can you change these ENV var names to 
OCP_USERNAME  from OPENSHIFT_USER
OCP_PASSWORD from OPENSHIFT_PASSWORD

because as per @durera comment on my PR [Check this comment](https://github.com/ibm-mas/ansible-devops/pull/187#discussion_r825983293) i have modified these env vars

multicloud-bootstrap/init.sh
```diff
@@ -0,0 +1,296 @@
+#!/bin/bash
+#
+# This is the init script that will call the individual Cloud specific script
+#
+
+## Inputs
+export CLOUD_TYPE=$1
```
Ashish-Mahawal1, line 7: as per @santoshjpawar this env var should be changed to 
CLUSTER_TYPE because roles are consuming same value as CLUSTER_TYPE
[Discussion thread](https://ibm-watson-iot.slack.com/archives/C031QACEAT1/p1647240145105549?thread_ts=1646900393.727049&cid=C031QACEAT1)

amitmangalvedkar: Done, I have made the changes

multicloud-bootstrap/init.sh
amitmangalvedkar: Done, I have made the changes

amitmangalvedkar:
> Please can you add a README.md into the multicloud-bootstrap directory that describes the content, what it does, how to use it etc. A good example is this one here: https://github.com/ibm-mas/ansible-devops/blob/master/pipelines/README.md ... we should have something like this for the multicloud-bootstrap content

Added `README.md` - in case you are ok with the content, could you please accept the PR?

durera (approved):
Nice :+1:

PR #184, opened 2022-03-09
[patch] Extra logic to recover from BAS install failures

durera:
The BAS operator isn't very good.  It fails to install often, this hack addresses one main cause of install failure with BAS: the `createcluster` job fails and the BAS operator cannot handle the failure.

### DIFF
--- a/ibm/mas_devops/roles/bas_install/tasks/main.yml
+++ b/ibm/mas_devops/roles/bas_install/tasks/main.yml
@@ -132,14 +132,75 @@
   no_log: "{{ bas_grafana_password is defined }}"
 
 
-# 7. Create BAS Full Deployment
+# 7. Create BAS AnalyticsProxy
 # -----------------------------------------------------------------------------
 - name: "Create BAS AnalyticsProxy"
   community.kubernetes.k8s:
     definition: "{{ lookup('template', 'templates/analyticsproxy.yaml') }}"
 
 
-# 8. Wait for AnalyticsProxy to be complete
+# 8. BAS Install Reliability Hack
+# -----------------------------------------------------------------------------
+# BAS is VERY unreliable
+# Instead of waiting for the AnalyticsProxy status to report success as we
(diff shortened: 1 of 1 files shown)


ibm/mas_devops/roles/bas_install/tasks/main.yml
```diff
@@ -132,27 +132,88 @@
...
+    kind: AnalyticsProxy
+  register: bas_cr_lookup
+  until:
+    - bas_cr_lookup.resources is defined and bas_cr_lookup.resources | length == 1
+    - bas_cr_lookup.resources[0].status is defined
+    - bas_cr_lookup.resources[0].status.phase is defined
+  retries: 15 # approx 15 minutes before we give up waiting for status.phase to be initialized
+  delay: 60 # 1 minute
+
+- name: "Pause for 20 minutes to wait for the createcluster job to complete"
```
andrercm, line 177: Any reason to not use a conditional waiting as we do in other places here?
Example: https://github.com/ibm-mas/ansible-devops/blob/master/ibm/mas_devops/roles/amqstreams/tasks/main.yaml#L72

andrercm (changes requested):

ibm/mas_devops/roles/bas_install/tasks/main.yml
durera: I couldn't think of a good way to make it work reliably given the odd way the job is managed by the BAS operator.  Iit not being there could mean either the job hasn't ran at all yet, or it has ran and completed successfully.  This seemed the safest/most reliable way to handle things.  

If the AnalyticsProxy isn't already in Ready state when we start then we wait 15 minutes, and if after 15 minutes this job exists we delete it.  After that we do the normal "wait until Ready" for the AP.

If the AnalyticsProxy is in Ready state already then none of this new code exexcutes (other than the initial check to decide whether to activate this hack).

PR #189, opened 2022-03-10
[minor] Add Workspace config generator

shajeena:
Implements a new role (`gencfg_workspace`) that is able to generate a Workspace custom resource and save the file to local disk.

Supports standard set of variables to configure the workspace:

- `mas_instance_id`
- `mas_workspace_id`
- `mas_workspace_name`

This allows us to remove the need for a user to pass in a pre-created workspace CR into the `mas_config_dir`.  Existing playbooks will be updated to take advantage of this capability, but a user does not need to use this, they can still pass a pre-created workspace custom resource directly to `suite_config`.  This just provides an alternative.


shajeena:
@durera - Can you please review the changes

ibm/mas_devops/roles/suite_install/tasks/airgap-catalogs.yml
```diff
@@ -0,0 +1,43 @@
+---
```
durera, line 1: - Anything airgap related should go into the mas_airgap extension that is being worked on.
- Catalog installation is performed in ocp_setup_mas_deps role already

ibm/mas_devops/roles/suite_install/templates/airgap-catalogs.yml.j2
```diff
@@ -0,0 +1,36 @@
+---
```
durera, line 1: - Anything airgap related should go into the mas_airgap extension that is being worked on.
- Catalog installation is performed in ocp_setup_mas_deps role already

ibm/mas_devops/roles/suite_config/tasks/main.yml
```diff
@@ -48,3 +73,13 @@
     state: present
     definition: "{{ lookup('template', item.path) }}"
   with_items: "{{ mas_config_files.files }}"
+
```
durera, line 76: If want to have a role create a workspace let's put it in a seperate new role please, e.g. `mas_workspace_create` ... this role's job is to apply configurations from a directory.  If one wants to use this to create a workspace they create the yaml for the workspace and pass the directory containing that file into the role (as we do today in daily FVT).

.idea/.gitignore
```diff
@@ -0,0 +1,3 @@
+# Default ignored files
```
durera, line 1: These .idea files should be added to .gitignore and removed from the repo :+1:

durera (changes requested):

ibm/mas_devops/roles/suite_install/tasks/airgap-catalogs.yml
shajeena: @durera - Thanks for the review 
Have removed the airgap extension from suite_install

ibm/mas_devops/roles/suite_install/templates/airgap-catalogs.yml.j2
shajeena: Have removed the airgap extension from suite_install

ibm/mas_devops/roles/suite_config/tasks/main.yml
shajeena: Created a new role suite_config_workspace and added the role in install_suite.yml . Please let me know if this is correct .
@santoshjpawar - FYI

.idea/.gitignore
shajeena: Wrong checkin :) , thanks removed it

shajeena:
@durera - can you please review comments addressed and approve the PR.Thanks

ibm/mas_devops/roles/suite_config/tasks/main.yml
```diff
@@ -1,17 +1,41 @@
...
+  when: mas_workspace_id is not defined or mas_workspace_id == ""
+  fail:
+    msg: "mas_workspace_id property is required"
+
 - name: "Fail if mas_config_dir is not provided"
   when: mas_config_dir is not defined or mas_config_dir == ""
   fail:
     msg: "mas_config_dir property is required"
 
+# Wait for Mongocfg CRD
```
durera, line 20: We can't assume the mongocfg is being applied here.  We do need something like this, but it needs to be more generic and able to handle the different xCfg resources that might (or might not) be passed into this role

durera (changes requested):
See review comments.  I'll need to make some adjustments to the larger pipelines to support the changes in this PR before it's merged.  Will look to do that tomorrow.

ibm/mas_devops/roles/suite_config/tasks/main.yml
shajeena: @santoshjpawar - Can you please on this comment.

durera (approved):

Issue #87, opened 2021-12-04
Support db2u standalone operator


durera:
After discussions with the team, it's been advixed that we use the standalone db2u operator instead of CloudPak for Data to meet our Db2 dependency.

The task here is to implement **alternative** support for the standalone operator in a new role (`db2u`) as an alternative to the existing `cp4d_db2wh` role.  

In the future we may eliminate the `cp4d_db2wh` role, but for now we just want to introduce this support as an alternative.  Deploying Db2 without CP4D will reduce install time and footprint.

The work has already been started in the db2u branch, however after my accident no-one picked the work up so it's remained in limbo for a number of weeks now.

The role needs to:
- install the operator (if it's not already installed)
- create a db2 instance
- not do anything with CP4D at all

PR #203, opened 2022-03-22
[minor] Add support for db2u operator

whitfiea:
Adds support for the db2u operator by adding a new `db2u` role. This will create a db2wh type database in the same way that the `cp4d_db2wh` role does but without the need to install cp4d. The configuration and options are the same as the cp4d-db2wh version as both use the same db2ucluster resource kind. The db2u operator is installed via ibm-common-services operand request.

This is for the enhancement in issue https://github.com/ibm-mas/ansible-devops/issues/87


ibm/mas_devops/playbooks/dependencies/install-db2.yml
```diff
@@ -0,0 +1,16 @@
+---
+- hosts: localhost
+  any_errors_fatal: true
+  vars:
```
durera, line 4: Can all of these vars be removed to align with the newer style of playbook where we pushed all the lookups/defaults into the role's defaults/main.yml so that they are all in one place inside the role?

Looing at the list of vars I think all of these are covered by existing entries in the role already.

durera (changes requested):

ibm/mas_devops/playbooks/dependencies/install-db2.yml
whitfiea: Removed the vars and updated the playbook dependencies doc to show what environment variables are required.

durera (approved):

PR #204, opened 2022-03-22
[minor] adding support for install mas with manual upgrade

CAIOFCP:
This PR contains changes to install MAS and Apps using Manual upgrade strategy as default. This is a requirement for any SRE environment, considering that any potential outage should be communicated to customers, which may also be true for other customers as well.

When installing MAS using Manual Upgrade Strategy, all ibm-common-services operators are automatically set to Manual in ibm-common-services namespace. For the cases where common services were previously installed in the cluster, their operators subscriptions will shows as Automatic but functioning as Manual.


CAIOFCP:
image

CAIOFCP:
image

CAIOFCP:
image

PR #205, opened 2022-03-22
[minor] Support manual upgrade approvals

CAIOFCP:
This PR contains changes to install MAS and Apps using Manual upgrade strategy as default. This is a requirement for any SRE environment, considering that any potential outage should be communicated to customers, which may also be true for other customers as well.

When installing MAS using Manual Upgrade Strategy, all ibm-common-services operators are automatically set to Manual in ibm-common-services namespace. For the cases where common services were previously installed in the cluster, their operators subscriptions will shows as Automatic but functioning as Manual.


CAIOFCP:
image

CAIOFCP:
image

CAIOFCP:
image

ibm/mas_devops/roles/suite_install/defaults/main.yml
```diff
@@ -7,3 +7,10 @@ mas_icr_cp: cp.icr.io/cp
 mas_entitlement_username: cp
 
 mas_icr_cpopen: icr.io/cpopen
+
+ibm_common_services_namespace: ibm-common-services
```
durera, line 11: Can you add a comment here explaining that this is what controls the install plans that are in scope in the manual approval codepath etc. So that it's obvious to future reader what this is doing here 👍

durera (changes requested):
See review comments, looks good overall ...

durera (dismissed):

durera (approved):

PR #224, opened 2022-03-30
[minor] Support GPU node in OCP on ROKS

caroazad:
- Added GPU node deployment to the ocp_provision role for roks. 
- Included environment variables OCP_PROVISION_GPU and GPU_WORKERS
- Updated docs for ocp_provision
- Tested with MVI installation (Images included)

**Created GPU worker pool**
Screen Shot 2022-03-29 at 10 14 52 PM

**GPU node is active in console**
Screen Shot 2022-03-29 at 10 18 11 PM

**Number of GPU displays in MVI application (2 X # of worker nodes)**
Screen Shot 2022-03-29 at 8 52 29 PM


ibm/mas_devops/roles/ocp_provision/tasks/providers/roks.yml
```diff
@@ -110,7 +110,61 @@
...
+      --name gpu \
+      --flavor mg4c.32x384.2xp100 \
+      --size-per-zone {{ gpu_workers }} \
+      --entitlement cloud_pak
+
+- name: "roks : Add GPU worker pool to zone"
+  when:
+    - ocp_provision_gpu is defined
+    - ocp_provision_gpu|bool  == true
+    - workerpool_lookup.rc == 1
```
whitfiea, line 149: It looks like we only do this task if the workerpool lookup earlier reported that it was already present, but it could be present but not added to the zone (due to an error after the worker-pool create). So next time it is called it doesn't add the worker to the zone. Is there harm in just calling `ibmcloud oc zone add` if it is already added?

ibm/mas_devops/playbooks/fullstack-roks.yml
```diff
@@ -220,27 +216,22 @@
...
+        storage:
+          size: 100Gi
+          storageClassName': ibmc-file-gold
+# add to mas_app_spec such as to configure the binding for Edge addon if needed when creating your own playbooks.
+
+- name: Configure Maximo Visual Inspection
+  import_playbook: mas/configure-app.yml
+  vars:
+    mas_app_id: visualinspection
+    mas_workspace_id: masdev
```
whitfiea, line 234: Should the workspace id be hardcoded to masdev?

whitfiea (changes requested):

ibm/mas_devops/playbooks/fullstack-roks.yml
caroazad: I think in the fullstack-roks.yml file it is assumed that the mas_workspace_id is masdev. If the user creates their own playbook, they can change the mas_workspace_id.  I am just using the same workspace id that was created and used by all the other application in this file.

ibm/mas_devops/roles/ocp_provision/tasks/providers/roks.yml
caroazad: Yes, there is an error if the worker-pool has already been added to the zone. I am going to add a part to check if it has been added or not before adding. Maybe that would solve this problem

ibm/mas_devops/playbooks/fullstack-roks.yml
whitfiea: oh yes I see what you mean. I didn't notice that in the compare view.

ibm/mas_devops/roles/ocp_provision/tasks/providers/roks.yml
caroazad: changed made and tested

whitfiea (changes requested):
Can we add the gpu_install role to docs index here https://github.com/ibm-mas/ansible-devops/blob/master/mkdocs.yml#L23

Also I assume that gpu nodes are required for MVI so there is little point (as it takes a long time to provision) to adding the install/cfg MVI to the sample pipeline: https://github.com/ibm-mas/ansible-devops/blob/master/pipelines/samples/sample-pipeline.yaml#L156.

whitfiea:
@caroazad yeah that makes sense to have it as a seperate issue as this PR just adds the gpu node install. I have opened issue https://github.com/ibm-mas/ansible-devops/issues/235 to add MVI to the sample pipeline

whitfiea (approved):

Issue #215, opened 2022-03-24
Shouldn't put server certificates into suite cfg resources


whitfiea:
The certificates section in a suite resource cfg should just contain the CA certificate (and any intermediate CA certificates) and not the server certificate. An example of this is for mongo here https://github.com/ibm-mas/ansible-devops/blob/master/ibm/mas_devops/roles/mongodb/templates/community/suite_mongocfg.yml#L43

The certificates section is to define what certificates are trusted and not to contain the actual server certificate that will likely expiry soon.

PR #225, opened 2022-03-30
[patch] Remove server cert from mongocfg

whitfiea:
Removes the server certificate from the mongocfg as only the CA certificate should be added, see issue #215 . Also removes the openssl template as it is not used anymore.


durera (approved):
LGTM

PR #187, opened 2022-03-10
[patch] AWS Support: mongodb, ocp_login, & ocp_setup_mas_deps

Ashish-Mahawal1:
PR includes changes in
Playbook :
`configure-ocp.yml`
Roles :
```
ocp_setup_mas
ocp_verify
ocp_login
mongodb
amqstream
```

PR #233, opened 2022-03-30
[patch] Fix breaking change in #187

durera:
- Looks like a really bad merge in #187.  Foundation services install was moved into main, but has been restored to the cert-manager sub-task in that PR.
- Also, makes the modification to `ClusterVersion` controllable via a new parameter `ocp_disable_upgrade`, which defaults to false.
- Also, moves predict into a layer 3 in the pipeline so that it installs after both Monitor and Manage have installed.


durera:
Got past the regression: image

PR #236, opened 2022-04-01
[major] Consolidate cluster specific provision/deprovision playbooks

whitfiea:
Consolidates the current provision and deprovision playbooks that are cluster specific (provision-roks, provision-quickburn, deprovision-roks and deprovision-quickburn) to just a generic playbook for provision and a generic playbook for deprovision.

Docs/exmaples/playbooks have been updated to just call provision.yml or deprovision.yml and to set the CLUSTER_TYPE env var that is required.

This is a major release as it removes the existing playbooks that will have been used and introduces new playbooks. Anyone that was using the previous playbooks need to move to use the new provision and/or deprovision playbooks and ensure that the environment variables (as outlined in the playbook readme) are set.

Tested both new playbooks on both ROKS and Quickburn.


durera (approved):
Nice, LGTM :+1:

PR #238, opened 2022-04-04
[patch] Sensitive terraform variables

vskhabani:
Openshift password was getting displayed multiple times in terraform logs. Added code changes to mask the value of openshift password as sensitive value.


whitfiea (changes requested):
Should we also be adding the sensitive tag to the openshift_password in ocs and portworx variables? (https://github.com/ibm-mas/ansible-devops/blob/master/multicloud-bootstrap/aws/ocp-terraform/portworx/variables.tf and https://github.com/ibm-mas/ansible-devops/blob/master/multicloud-bootstrap/aws/ocp-terraform/ocs/variables.tf)

Also shouldn't the openshift_token be also classed as sensitive as that is the same as a password?

vskhabani:
Hi @whitfiea , with current changes I can see the openshift_password value getting treated as sensitive value in OCS as well and it is not getting printed anywhere in terraform apply. 
Attaching logs of terraform apply 
[provisioning-log-aws.txt](https://github.com/ibm-mas/ansible-devops/files/8408142/provisioning-log-aws.txt)

whitfiea:
@vskhabani looks like it must inherit the settings on the var as it isn't in the log as you said, even though ocs is used. What about adding sensitve to the openshift_token? As this is essentially the same as a password.

vskhabani:
Hi @whitfiea I have removed all unwanted code and also made other variables as sensitive.

whitfiea (approved):

PR #243, opened 2022-04-06
Azure support for OCP cluster deployment

santoshjpawar:
Updated bootstrap process to support deployments on Azure. This is only for the bootstrap process changes.


whitfiea (changes requested):
Looks like there is a random stackdump included: multicloud-bootstrap/azure/jq.exe.stackdump I assume that was a mistake?

santoshjpawar:
> Looks like there is a random stackdump included: multicloud-bootstrap/azure/jq.exe.stackdump I assume that was a mistake?

Thanks. Removed the file.

whitfiea (approved):

PR #246, opened 2022-04-06
Add variable for GPU worker pool name

caroazad:
I was working with the MVI team and it looks like we need to have a variable that can be set by the users for the gpu worker pool name they are using since they may not be using "gpu" or could also have a naming convention. I have made the necessary changes with the default name set to "gpu" but there will be the option to provide a name as well in the environment variables. Thank you


ibm/mas_devops/roles/ocp_provision/README.md
```diff
@@ -31,6 +31,12 @@ Flag that determines if GPU worker nodes should be added during cluster creation
 - Environment Variable: `OCP_PROVISION_GPU`
 - Default Value: `false`
 
+### gpu_wpool_name
```
whitfiea, line 34: A bit of a nit-pick but could we call the var `gpu_workerpool_name` rather than `gpu_wpool_name`? I don't think we have shorted names like that in other places. Environment variable name would change aswell

ibm/mas_devops/roles/ocp_provision/tasks/providers/roks.yml
```diff
@@ -112,11 +112,21 @@
...
 # 8. Add a GPU worker node if ocp_provision_gpu is true
 # -----------------------------------------------------------------------------
+- name: "roks : Debug Information for GPU worker"
+  when:
+    - ocp_provision_gpu is defined
+    - ocp_provision_gpu|bool == true
+  debug:
+    msg:
+      - "GPU woker pool name ................. {{ gpu_wpool_name }}"
+      - "GPU workers .................. {{ gpu_workers }}"
```
whitfiea, line 122: Can you add extra dots to this line so it lines up with the rest, see https://github.com/ibm-mas/ansible-devops/blob/master/CONTRIBUTING.md#align-debug-messages

ibm/mas_devops/roles/ocp_provision/README.md
```diff
@@ -31,6 +31,12 @@ Flag that determines if GPU worker nodes should be added during cluster creation
...
+The name of the gpu worker pool to added to or modify in the cluster. If already existing, use the existing name to avoid recreating another gpu worker pool unless that is the goal. 
+
+- Environment Variable: `GPU_WPOOL_NAME`
+- Default Value: `gpu`
+
 ### gpu_workers
 The number of GPU worker nodes to deploy in the cluster. Depends on `ocp_provision_gpu` and is currently only set up for ROKS clusters.
```
whitfiea, line 41: Could we add the flavour (mg4c.32x384.2xp100), to this description, that these worker nodes would be so users don't have to look in the code to see (or see after provision)

whitfiea (changes requested):

ibm/mas_devops/roles/ocp_provision/README.md
caroazad: okay, I will do that. :)

ibm/mas_devops/roles/ocp_provision/tasks/providers/roks.yml
caroazad: Thanks for sharing the document. very useful :)

ibm/mas_devops/roles/ocp_provision/README.md
caroazad: done

whitfiea (approved):

PR #251, opened 2022-04-11
[minor] Add secure route for cp4d web client when using CIS

whitfiea:
Adds a new secure route called `suite-cp4d-route`, during `suite_dns` role, in the cp4d namespace that is setup using the CUSTOM_DOMAIN set during `suite_dns` and creates the certificates using the CUSTOM_CLUSTER_ISSUER set during `suite_dns`. The `suite_dns` role is also updated to add a new DNS entry of `cp4d` (aslong as the DNS provider is CIS). If you set DNS entries via a different mechanism then a new entry should be added to DNS so that cp4d.{{mad_domain}} can be resolved to this route.

The secure route will reencrypt the connection to the CP4D web client (nginx) service. If CP4D has been setup with a custom cert then the route will continue to work as it will get the custom certificate from the correct CP4D secret.

A variable is also available to control if the route is created or not. This is for the case where you have setup a custom domain and use a custom cluster issuer but CP4D is not installed on this OCP cluster. In this case it is not possible to configure the route on the OCP cluster you are installing MAS on. The flag is called `configure_suite_cp4d_route` (as a playbook var) or `CONFIGURE_SUITE_CP4D_ROUTE` (as a env var) and defaults to true.

The end result is that https://cp4d.{{your_custom_mas_domain}} will be accessible using a certificate from your custom cluster issuer.

This has been tested using IBM CIS with a custom domain, and CP4D on the same OCP cluster.


whitfiea:
@durera The main path that this new setup will be used is when you have a custom domain and custom cluster issuer that uses CIS. The way to create this new route is to create new certificates using that custom cluster issuer but this is only setup during the suite_dns role that is part of the install-suite playbook. It does feel more logically to place this in the cp4d role but at the time that is run then it is unlikely that we will have a custom cluster issuer that has been setup. That is why it was added to the suite_install role as we have all the pieces available to setup the secure route.

whitfiea:
@durera I moved the logic into the suite_dns role now (previously in suite_install). I tested that it still works.

durera (approved):

PR #263, opened 2022-04-15
[minor] updated createUiDefinition.json file

vskhabani:
Updated createui json


whitfiea (changes requested):
Can you add more detail to the title and description of the PR that outlines what this is adding and why it is a minor commit?

Issue #49, opened 2021-10-29
Support Predict deployment and dependencies


alequint:
To complete support for Predict we actually need to complete support for Manage components.

Today in the code we have this:

```yaml
---
# Default application spec for Manage
mas_app_ws_spec:
  bindings:
    jdbc: "{{ mas_appws_jdbc_binding | default( 'system' , true) }}"
  components: "{{ mas_appws_components | default({'base': {'version': 'latest'}}, true) }}"
  settings:
    db:
      dbSchema: "{{ db2_schema }}"
      maxinst:
        demodata: "{{ manage_demo_data | bool }}"
        db2Vargraphic: true
        indexSpace: MAXINDEX
        tableSpace: MAXDATA
        bypassUpgradeVersionCheck: false
```

However, because our playbook strategy is to drive configuration via environment variables, there is no way to leverage this capability today as it requires settint vars specifically in the playbook; which we don't want to do becuase we want the config to be passed in via env vars.

To solve this, let's support something like this:

```
export MAS_APPWS_COMPONENTS=base=latest,health=latest
ansible-playbook playbooks/mas/configure-app.yml
```

Inside the `suite_app_configure` role we should have a new plugin that can parse name:value pairs in this env var into the object that we need to pass to the workspace cr spec.  This would be used in the defaults/main.yml file to set the default value for `mas_appws_components`, with a default if the env var is not set being `omit` (ie don't define the variable at all)

Then the pipeline itself will need updated to have a new param:

```yaml
    - name: mas_appws_components_manage
      type: string
      description: Manage components to configure in the workspace
      default: "base=latest,health=latest"
```

and also ensure this param is passed to the manage configure workspace step in the pipeline:

```yaml
    # 16.2 Configure Manage workspace
    - name: cfg-manage
      params:
        - name: junit_suite_name
          value: app-cfg-manage
        - name: mas_app_id
          value: manage
        - name: mas_workspace_id
          value: "$(params.mas_workspace_id)"
        - name: mas_appws_components
          value: "$(params.mas_appws_components_manage)"
      taskRef:
        name: mas-devops-configure-app
        kind: ClusterTask
```

Hopefully that gives a good idea of what we need here.

whitfiea:
@durera Just trying to parse this issue and what is needed. I am struggling with the need for the plugin :
> Inside the suite_app_configure role we should have a new plugin that can parse name:value pairs in this env var into the object that we need to pass to the workspace cr spec.

Why would we need to support 
[...]

PR #267, opened 2022-04-19
[major] Change MAS_APPWS_COMPONENTS variable format

whitfiea:
This PR implements issue #49 by changing to format of the MAS_APPWS_COMPONENTS variable from yaml to key=value pairs separated by commas. The defaults are also changed so that if nothing is set for MAS_APPWS_COMPONENTS then no components are installed in the workspace for the applications that support it. As of writing this PR that is only Manage and Health that these defaults. After this PR is merged then MAS_APPWS_COMPONENTS needs to be set prior to configuring Manage or Health.

For Manage only:
```
export MAS_APPWS_COMPONENTS="base=latest"
ansible-playbook playbooks/mas/configure-app.yml
```

For Manage with Health:
```
export MAS_APPWS_COMPONENTS="base=latest,health=latest"
ansible-playbook playbooks/mas/configure-app.yml
```

For Health standalone:
```
export MAS_APP_ID=health
export MAS_APPWS_COMPONENTS="health=latest"
ansible-playbook playbooks/mas/configure-app.yml
```

The changes include a new filter to take the new format and change it to the yaml expected by the workspace CR. This has been tested in the pipeline to ensure that Manage can still be installed ok.


durera (dismissed):
As discussed in Slack ...

durera (approved):

PR #269, opened 2022-04-19
[minor] Added ability to set annotations onto suite CR

padmankosalaram:
**Change details **: 
Below annotation block needs to be added to suite CR in metadata section
```
metadata:
  annotations:
    mas.ibm.com/hyperscalerProvider: aws
    mas.ibm.com/hyperscalerFormat: saas 
    mas.ibm.com/hyperscalerChannel: ibm
```
To set this annotation block, we need an environment **MAS_ANNOTATIONS** variable in below format

``` 
export MAS_ANNOTATIONS="mas.ibm.com/hyperscalerProvider=aws:mas.ibm.com/hyperscalerFormat=saas:mas.ibm.com/hyperscalerChannel=ibm" 

```

**Implementation details**

Created below filter plugin which accepts the above annotation parameter and append to the CR definition
` ibm/mas_devops/roles/suite_install/filter_plugins/filters.py`


whitfiea (changes requested):
For the filters we have these are normally placed in https://github.com/ibm-mas/ansible-devops/blob/master/ibm/mas_devops/plugins/filter/filters.py rather than at the role level. If we then need to use it in other roles then we have access to it.

I am not sure that the addAnnotation filter works correctly when annotations (either the same or different) are already set. This means that the filter is not idempotent and so will give you different results if you run the task 1 vs 2 (or more) times. So I think the filter needs to check if annotations in metadata exists already or not, and then only add annotations if an annoatation key doesn't already exist. If the key exists then we need to check the value is the one expected (for the case where the value is changed but not the key).

You could probably this in jinga2 in the suite yaml template and use a filter that gives you a list of the key=value pairs in the MAS_ANNOTATIONS env var.

whitfiea (approved):

Issue #239, opened 2022-04-04
Enhance CP4D playbooks to fix critical performance issues


lokeshs09:
Opening an issue here make the necessary fixes or enhancements to the cp4d playbooks on anisble-devops. The recent breakdown, blockers on IVT10 and IVT11 were caused by 2 main factors. 

- auto updates to cp4d 4.0.7. 
- Performace issue caused by storage class

Details of the issue is been documented here: https://github.ibm.com/PrivateCloud-analytics/CPD-Quality/issues/2326

`Copying Sriram's recommendations here:`

**Summary of the issues:**

1)  Performance issue: - suspected storage IOPs (with ibmc-file-gold-gid) or even perhaps connection leaks - while one cluster perfoms better after scale out/up, the second cluster still has problems
 (@rahul-shinge & @kvstumph  reviewing the second cluster)

2) Issue with "ibm-operator-catalog" latest in use - this can cause uncontrolled automatic upgrades & will be hard for CPD to support if there are an arbitrary mix-n-match of versions (including Cloud Pak Foundational Services version)

---

**(future) Action Items:**

1)  Right storage selection (especially on IBM Cloud) to improve reliability   

When provisioning Ibmcpd CR -  add `zenCoreMetadbStorageClass`  to point to a block storage class

```
apiVersion: cpd.ibm.com/v1
kind: Ibmcpd
metadata:
  name: ibmcpd-cr
  namespace: cpd-instance
spec:
  license:
    accept: true
    license: Enterprise
  storageClass:  <rwx-storage-such-as-ibmc-file-gold-gid> 
 zenCoreMetadbStorageClass:  <block-storage>
```

2) Validating performance of the available storage classes
 CPD  now also has tools (published to open source) to measure/benchmark target storage
See: https://github.com/IBM/k8s-storage-perf

3)   Freeze the CP4D version - so it does not get randomly upgraded whenever any refresh happens  for stability - and important to be on a  “validated” version combination

using fixed catalog sources (image digests) instead of ibm-operator-catalog
https://www.ibm.com/docs/en/cloud-paks/cp-data/4.0?topic=ccs-creating-catalog-sources-that-pull-specific-versions-images-from-entitled-registry

CPD is  introducing additional automation  to reduce complexity of installs and upgrades while pinning versions:  https://github.ibm.com/PrivateCloud/olm-utils

[...]

PR #273, opened 2022-04-20
[patch] - fix cpd storage class

andrercm:
This PR rollbacks the change to use `ibmc-file-gold-gid` in both storage classes that cpd uses. Instead, now we'll again use `ibmc-file-gold-gid` for the primary storage class and `ibmc-block-gold` for zen metadb. This was discussed and agreed after the recent cpd outage we had in march on ivt11, the recommendation came from cpd support team:

https://github.ibm.com/PrivateCloud-analytics/CPD-Quality/issues/2326

In addition, this PR also updates Watson services to latest patch available.

### DIFF
--- a/ibm/mas_devops/roles/cp4d_install/defaults/main.yml
+++ b/ibm/mas_devops/roles/cp4d_install/defaults/main.yml
@@ -10,12 +10,9 @@ cpd_registry_user: cp
 # This is only used for cpd35, we have to use cp4d_hack_worker_nodes to set up the global image pull secret needed by cpd40
 cpd_entitlement_key: "{{ lookup('env', 'CPD_ENTITLEMENT_KEY') }}"
 
-# Both CP4D storage classes have the same storage requirements as per recommendation from cpd support team
-# so we set storageClass and zenCoreMetadbStorageClass to "{{ cpd_storage_class }}"
-# https://www.ibm.com/docs/en/cloud-paks/cp-data/4.0?topic=requirements-storage
-#
 # If this changes in the future we will expose a seperate variable for each.
 cpd_storage_class: "{{ lookup('env', 'CPD_STORAGE_CLASS') }}"
+cpd_metadb_block_storage_class: "{{ lookup('env', 'CPD_METADB_BLOCK_STORAGE_CLASS') }}"
 
 # --- MAS channel control ---------------------------------------------------------------------------------------------
 # Temporary solution: this will be used to determine whether to install CPD v4 or v3.5, if MAS channels from the below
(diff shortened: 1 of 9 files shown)


ibm/mas_devops/roles/cp4d_install/defaults/main.yml
```diff
@@ -10,12 +10,9 @@ cpd_registry_user: cp
 # This is only used for cpd35, we have to use cp4d_hack_worker_nodes to set up the global image pull secret needed by cpd40
 cpd_entitlement_key: "{{ lookup('env', 'CPD_ENTITLEMENT_KEY') }}"
 
-# Both CP4D storage classes have the same storage requirements as per recommendation from cpd support team
-# so we set storageClass and zenCoreMetadbStorageClass to "{{ cpd_storage_class }}"
-# https://www.ibm.com/docs/en/cloud-paks/cp-data/4.0?topic=requirements-storage
-#
 # If this changes in the future we will expose a seperate variable for each.
 cpd_storage_class: "{{ lookup('env', 'CPD_STORAGE_CLASS') }}"
+cpd_metadb_block_storage_class: "{{ lookup('env', 'CPD_METADB_BLOCK_STORAGE_CLASS') }}"
```
whitfiea, line 15: Can we have this new env var default to whatever CPD_STORAGE_CLASS is if CPD_METADB_BLOCK_STORAGE_CLASS is not set? Then at least it is backwards compatible. It might not be the recommended for a production CP4D but it would still be a supported setup I believe.

whitfiea (changes requested):
Added a comment, plus is this PR part of issue https://github.com/ibm-mas/ansible-devops/issues/239 as it isn't mentioned or linked to that issue?

ibm/mas_devops/roles/cp4d_install/defaults/main.yml
andrercm: I can include a default value for both vars but it would only be applicable for ROKS, for Fyre/Quickburn it woudn't work as they have different storage classes.
I think that's the reason why we didn't specified default values and instead added a check to ensure both are set in the `main.yml` task.

whitfiea: @andrercm For the default value I meant if CPD_METADB_BLOCK_STORAGE_CLASS is not set then lookup CPD_STORAGE_CLASS instead, rather than set a specific storage class default. This way users that have just CPD_STORAGE_CLASS set will still continue to work but can set CPD_METADB_BLOCK_STORAGE_CLASS if required.

whitfiea (approved):
I updated this to a major, as we are introducing a new required var to an existing role. So this would break users if upgraded and the var is not added.

whitfiea (approved):

PR #274, opened 2022-04-22
Added "azure" as supported_cluster_types list. 

vaibhavkulkarniibm:


whitfiea (changes requested):
This doesn't seem to add anything other than azure to a list. I assume that with azure you would log into the OCP cluster using the generic login that is in place, in which case it shouldn't be in the list. You could also argue that aws shouldn't be on the list and shoud log in with just the ocp token using the geneic way.

vaibhavkulkarniibm:
> This doesn't seem to add anything other than azure to a list. I assume that with azure you would log into the OCP cluster using the generic login that is in place, in which case it shouldn't be in the list. You could also argue that aws shouldn't be on the list and shoud log in with just the ocp token using the geneic way.

hello @whitfiea, 
yes we're doing generic login to ocp. But "azure" is needed in list due to dependency of following when condition. 
https://github.com/ibm-mas/ansible-devops/blob/44c1bfc750894f17e29fa62c41ed09c7011e75c6/ibm/mas_devops/roles/ocp_login/tasks/main.yml#L9-L16

whitfiea:
@vaibhavkulkarniibm "azure" needs to **not** be in the list for that when condition to be true. When that condition is true then it will then check that ocp_token and ocp_server have been set which is needed for the generic login.

vaibhavkulkarniibm:
Closing the PR, since changes not required for generic login

PR #278, opened 2022-04-27
[patch] Fixes for gencfg_uds and gencfg_sls

shubhammalvankar:
This update:
- Consolidates the naming convention for SLS variables, the handful of  `slscfg_xxx` variables are now `sls_xxx`
- Addresses defect introduced in previous update regarding support for `UDS_TLS_CERT_LOCAL_FILE`
- Moves default logic previously added into playbooks into the roles
- Updates documentation to address numerous gaps

## Information about the defect
The default in the gencfg_uds `uds_tls_crt: "{{ lookup('file', UDS_TLS_CERT_LOCAL_FILE) | regex_findall('(-----BEGIN .+?-----(?s).+?-----END .+?-----)', multiline=True, ignorecase=True) }}"` is causing the FVT to fail at the task `Lookup Certificate for UDS (custom)` as it is combining all the defaults from both the roles uds_install and gencfg_uds in the playbook. Hence it is not able to find the variable UDS_TLS_CERT_LOCAL_FILE which is needed for gencfg_uds and not uds_install.

Here the fix added is to move the variables from the role defaults to the playbook.

Error message seen:

```
TASK [ibm.mas_devops.uds_install : Lookup Certificate for UDS (custom)] *******************************************************************************
fatal: [localhost]: FAILED! => {"msg": "The conditional check 'uds_tls_crt is not defined' failed. The error was: error while evaluating conditional (uds_tls_crt is not defined): {{ lookup('file', UDS_TLS_CERT_LOCAL_FILE) | regex_findall('(-----BEGIN .+?-----(?s).+?-----END .+?-----)', multiline=True, ignorecase=True) }}: 'UDS_TLS_CERT_LOCAL_FILE' is undefined\n\nThe error appears to be in '/root/.ansible/collections/ansible_collections/ibm/mas_devops/roles/uds_install/tasks/udscfg.yml': line 110, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Lookup Certificate for UDS (custom)\n ^ here\n"}
```


durera (changes requested):
What is the reason behind this change?  

This is the opposite of what our approach to defaults management is.  When defaults are in the role, they work wherever the role is used, when used in a playbook the default logic needs to be reproduced everywhere?

shubhammalvankar:
> What is the reason behind this change?
> 
> This is the opposite of what our approach to defaults management is. When defaults are in the role, they work wherever the role is used, when used in a playbook the default logic needs to be reproduced everywhere?

The current approach was failing since the gencfg_uds role had a default ENV variable UDS_TLS_CERT_LOCAL_FILE which is only needed for existing UDS but in this case, since both uds_install and gencfg_uds roles were called in the playbook install-uds. It was overwriting the variable and causing failure while looking for the certs in the uds_install role.

Hence @whitfiea suggested to set the variables in the playbook itself rather than using them in defaults.

durera:
> > What is the reason behind this change?
> > This is the opposite of what our approach to defaults management is. When defaults are in the role, they work wherever the role is used, when used in a playbook the default logic needs to be reproduced everywhere?
> 
> The current approach was failing since the gencfg_uds role had a default ENV variable UDS_TLS_CERT_LOCAL_FILE which is only needed for existing UDS but in this case, since both uds_install and gencfg_uds roles were called in the playbook install-uds. It was overwriting the variable and causing failure while looking for the certs in the uds_install role.
> 
> Hence @whitfiea suggested to set the variables in the playbook itself rather than using them in defaults.

[Ansible variable precedence](https://docs.ansible.com/ansible/latest/user_guide/playbooks_variables.html#understanding-variable-precedence) puts role defaults pretty much as the last thing that takes precedence.

In a playbook that calls uds_install first, and then gencfg_uds afterwards.  The code in uds_install should be setting the `uds_tls_crt` fact, which will ensure that the default for that fact in the `gencfg_uds` role is overridden.  The role default cannot override the existing fact that was set in the earlier role.  We've been using this in daily FVT for a while now, as you sure this is what's going on .. the same approach is used with other install/gencfg role combinations, e.g. `gencfg_jdbc` supports the exact same model with environment variables in the defaults, which are not used because we set the variable via a fact in the role that runs before.

whitfiea:
@durera The PR text is missing the details on the error seen which would provide some context to the problem. Also the default in the gencfg_uds `uds_tls_crt: "{{ lookup('file', UDS_TLS_CERT_LOCAL_FILE) | regex_findall('(-----BEGIN .+?-----(?s).+?-----END .+?-----)', multiline=True, ignorecase=True) }}"` has been recently added and is now causing the failure in FVT.

>The code in uds_install should be setting the uds_tls_crt fact, which will ensure that the default for that fact in the gencfg_uds role is overridden

That is true but the issue is that the role doesn't get to that point when it fails. When the playbook runs all the default vars for each role is merged before any roles/tasks are executed, so the `uds_tls_crt` that is defined in the `gencfg_uds` roles defaults/main.yaml is defined when the playbook runs. So when it gets to the first role of `uds_install` we get to this task https://github.com/ibm-mas/ansible-devops/blob/161711691b32583c9121a73494bd856a913eb415/ibm/mas_devops/roles/uds_install/tasks/udscfg.yml#L111 which fails as it tries to resolve as `uds_tls_crt ` as `uds_tls_crt: "{{ lookup('file', UDS_TLS_CERT_LOCAL_FILE) | regex_findall('(-----BEGIN .+?-----(?s).+?-----END .+?-----)', multiline=True, ignorecase=True) }}"` which fails as the lookup doesn't work. It isn't overridden as no set_fact has been called yet in `uds_install` (the one [here](https://github.com/ibm-mas/ansible-devops/blob/161711691b32583c9121a73494bd856a913eb415/ibm/mas_devops/roles/uds_install/tasks/udscfg.yml#L100) has been skipped due to the condition, and the next set_fact is after the failing task [here](https://github.com/ibm-mas/ansible-devops/blob/161711691b32583c9121a73494bd856a913eb415/ibm/mas_devops/roles/uds_install/tasks/udscfg.yml#L126)

We could work around it by not checking if uds_tls_crt is defined or not (choosing some other var to check) and just letting uds_install set its facts.

shubhammalvankar:
Added [patch] to the commit message @whitfiea

whitfiea:
@shubhammalvankar can you edit the first comment in this PR to have more than just "Moved the gencfg_uds role variables to the playbook gencfg-uds". PRs should describe why you are doing the change and not just what the change is. Please include the error messages seen etc.

shubhammalvankar:
> @shubhammalvankar can you edit the first comment in this PR to have more than just "Moved the gencfg_uds role variables to the playbook gencfg-uds". PRs should describe why you are doing the change and not just what the change is. Please include the error messages seen etc.

@whitfiea The comment has been edited to include the error and reason.

whitfiea (approved):
Although it is adding vars into the playbook I don't think we have a better solution.

whitfiea (changes requested):
@shubhammalvankar After speaking with David about this change, it would be better to move the vars out of the playbook and into the defaults, but to be used differently. If you set the defaults in gencfg_uds to be like this:

```
uds_tls_crt_local_file_path: "{{ lookup('env', 'UDS_TLS_CERT_LOCAL_FILE_PATH') }}"
uds_tls_crt: "{{ lookup('env', 'UDS_TLS_CRT') }}"
uds_endpoint_url: "{{ lookup('env', 'UDS_ENDPOINT_URL') }}"
uds_api_key: "{{ lookup('env', 'UDS_API_KEY') }}"
```

then you can remove the vars from the gencfg-uds playbook. The `gencfg_uds` role can then check if uds_tls_crt_local_file_path is defined and not "" and use that to then set uds_tls_crt with the role. You would also need to update anywhere in the `install_uds` role that that does a `is uds_tls_crt defined` and add a `and uds_tls_crt != ""` to those conditions.

This then needs to be tested in the pipeline that uses the install_uds playbook.

shubhammalvankar:
@whitfiea I have moved the variables from the playbook to the role defaults 
```
uds_tls_crt_local_file_path: "{{ lookup('env', 'UDS_TLS_CERT_LOCAL_FILE_PATH') }}"
uds_endpoint_url: "{{ lookup('env', 'UDS_ENDPOINT_URL') }}"
uds_api_key: "{{ lookup('env', 'UDS_API_KEY') }}"
```

Added a task `- name: "Set uds_tls_crt fact for BasCfg when using existing UDS"` which will run `when: uds_tls_crt is not defined and uds_tls_crt_local_file_path is defined and uds_tls_crt_local_file_path != ""`.

This way we don't have to set `uds_tls_crt` in defaults, rather just set it as a fact when it's not defined and if `uds_tls_crt_local_file_path` is set.

ibm/mas_devops/roles/gencfg_sls/defaults/main.yml
```diff
@@ -5,4 +5,6 @@ mas_config_dir: "{{ lookup('env', 'MAS_CONFIG_DIR') }}"
 
 slscfg_url: "{{ lookup('env', 'SLSCFG_URL') }}"
 slscfg_registration_key: "{{ lookup('env', 'SLS_REGISTRATION_KEY') }}"
-slscfg_tls_crt: "{{ lookup('file', SLS_TLS_CERT_LOCAL_FILE) }}"
+
+sls_tls_crt_local_file_path: "{{ lookup('env', 'SLS_TLS_CERT_LOCAL_FILE_PATH') }}"
+slscfg_tls_crt: "{{ lookup('file', sls_tls_crt_local_file_path) }}"
```
whitfiea, line 10: Can we remove this like we do in gencfg_uds and have it set in a fact at the start of the role. Otherwise we have two different mechanisms for similar things.

whitfiea (changes requested):
Added one comment. Plus have you tested this change using the install_uds playbook?

whitfiea (approved):

whitfiea (approved):

whitfiea:
changes tested and working in pipeline

durera (approved):

durera (approved):

PR #280, opened 2022-04-28
[patch] Added MVI and its dependencies in the sample pipeline

caroazad:
Included install-gpu in the pipeline
Included MVI installation and configuration in the pipeline

I still need to add the channels for MVI in the sample files. I will meet with Dominique to get them soon.


whitfiea:
@caroazad is this ready for review now, as you mentioned needing to add the channels but it looks like that is done?

caroazad:
> @caroazad is this ready for review now, as you mentioned needing to add the channels but it looks like that is done?

Yes, it is done and ready for review

caroazad:
I was wondering if I should add the ocp_provision_gpu var and others in the sample settings for the customers. Just in case they want to add gpu provisioning in their pipeline. What are your thoughts?

whitfiea:
@caroazad I think we can leave out the **ocp_provision_gpu** as it is an option that is mentioned in the readme for the role, so anyone looking at adding it to a pipeline would see that setting.

PR #282, opened 2022-04-29
[patch] fix segmentKey shows the omit_place_holder instead of omitted in role gencfg_uds

Yolandabjyf:
Fix issue#277

segmentKey is now only set in the BAScfg CR when mas_segment_key is set.


whitfiea (approved):
Thanks for the change, looks good to me

Issue #261, opened 2022-04-14
Rename playbooks so these can be called via the collection


whitfiea:
In ansible 2.11 you can now call the playbooks using the collection syntax similar to how you call roles. See https://docs.ansible.com/ansible/latest/user_guide/collections_using.html#using-a-playbook-from-a-collection for more details. The problem with our playbooks is all of them have a dash (-) in the filename and as per that link that is not supported.

If we rename the playbooks to use underscore rather than dash then users (assuming those users have ansible 2.11 or above) will be able to call the playbooks like this:
```
ansible-playbook ibm.mas_devops.lite_core_quickburn 
```

durera:
Updates merged into master branch

PR #285, opened 2022-04-30
[major] Initial one-click installer support

durera:
## Overview 
- #261 
- First release to support https://github.com/ibm-mas/installer
- Restructuring of playbooks to create clear seperation between provisioning and installation to better support "install anywhere"
    - onclick_x playbooks are the major playbooks that replace lite/fullstack
    - ocp_roks_x are playbooks for managing OCP in ROKS
    - ocp_fyre_x are playbooks for managing OCP in FYRE
    - The equivalent of the old fullstack/lite playbook is to chain together your choice of provision playbook with the - now generic - oneclick playbook
    - One playbook per application only
- Separate image registry capacity upgrade into it's own role (that makes it clear this **only** works for ROKS)
- Numerous documentation updates
- Bug fix for db2u role:  Code did not wait for CRD to be available before creating the custom resource
- Removal of support for CP4D Db2, superceeded by Db2Universal operator (same thing, but smaller footprint, faster install, less problematic versioning system)
- Updates Tekton Tasks to use new playbook structure, there is now a clear 1:1 mapping between Ansible roles and Tekton Tasks
- Add Db2u ClusterTask
- Remove sample pipeline (this is now in https://github.com/ibm-mas/installer)

### Single role playbooks are gone
These were an artifact of "my first ansible collection", and the approach ended up being copied everywhere, becoming the defacto model.  These are all dead.

To run a single role do this instead: `ansible localhost -m include_role -a name=ibm.mas_devops.ocp_setup_ocs`.  If you want to run a single role as a playbook (to get the additional benefits that brings):

```export ROLE_NAME=ocp_setup_ocs
ansible-playbook ibm.mas_devops.run_role
```

All playbook docs have now been merged into the role docs eliminating one of the more confusing natures of our docs that almost everything we supported had two pages that had almost identical content (but naturally they weren't both kept in sync and often conflicted one another).

### What happened to the fullstack and lite playbooks?
They are also gone.  These became problematic as there was 90% duplicated code in them.  They are replaced with a one-click install friendly seperation of provisioning and MAS installation.  This allows one "install mas core" playbook to be used regardless of whether it's being installed on ROKS, Azure, AWS, or any random datacenter.

Each application will offer a single playbook that acts as an extension of the core playbook, rather than containing everything it contained plus a couple of extra lines to install 1 application.

Thus the model, if you want to use playbooks directly, now becomes something like this, where the playbooks act as building blocks:
```export ALL_YOUR_VARS
ansible-playbook ibm.mas_devops.ocp_roks_provision
ansible-playbook ibm.mas_devops.oneclick_install_suite
ansible-playbook ibm.mas_devops.oneclick_add_manage
ansible-playbook ibm.mas_devops.oneclick_add_iot
ansible-playbook ibm.mas_devops.oneclick_add_<yourapphere>
```

### Db2 is no longer tied to CP4D
Db2 Universal Operator support is where it's at.  It's the same operator you get from CP4D, just without any of the baggage or the versioning problems CP4D brings with it.

### Intelligent storage class defaults
Storage class variables are now almost universally optional when deploying in ROKS, when not defined the role will look for the ROKS storage classses and use the right ones if they are available.  Some roles also do this for OCS storage classes (where we know what the right OCS storage class is).  It's an expectation going forward all use of storage classes will work this way.  We also want this to be expanded ... e.g. ideally for everything that needs a storage class we would know what the right OCS storage class to use is as well.

### Tekton ClusterTasks Revamped
They are no longer ClusterTasks .. they are simply Tasks (namespace scoped).  They also now directly align 1:1 with roles in the ansible collection.  If there is a role named `my_role`, there will be a task named `mas-devops-my-role`.  The params from the tasks will be updated to align with the variable names for the roles and the role documentation will be updated to cover the associated Tekton Task (and this will be a requirement for new roles going forward too)


durera:
More PRs will follow to deliver the rest of v10, want to get this into master so that all other PRs are based on the v10 redesign.

PR #290, opened 2022-05-05
[patch] update mkdocs.yml to include new roles in the nav panel

cuddlyporcupine:


durera:
This is covered in #292 :+1:

PR #291, opened 2022-05-05
Change gpu_install to nvidia_install

caroazad:
Changes to avoid confusion since the role is installing nvidia operators and not the gpu node


caroazad:
It looks like the naming changes have been made in a refactoring to come in the future so I will go ahead and close this PR and branch

PR #303, opened 2022-05-11
[patch] Increase SBO timeout

alequint:
Increasing number of attemps from 20 to 30

That is because operator took more than 20 minutes to be installed, which failed pipeline:

image

... and looking some time after,  Installplan instance was good and SBO overall has been installed successfully:

image

image


durera (approved):

durera (changes requested):
Can you fix the comment on the line below as well please.

```
  retries: 30
  delay: 60 # Retry for approx 20 minutes (60s * 20 attempts) before giving up
```

durera (approved):

PR #309, opened 2022-05-20
[minor] Update suite application roles to support Optimizer application

alequint:
- Update suite-app-install
- Update suite-app-config
- Add new attribute in `suite-app-config` to specify installation mode of Optimizer

Pending (this is a work in progress)
- [x] Update CR with proper installation mode (to be confirmed by Yuan Yuan)
- [x] Create OneClick playbook for MSO and Optimizer
- [x] Test
- [x] Second Test After PR review (in progress in cluster `alefq2505`); Update: test completed (note Optimizer successfully installed and not MSO):

image


ibm/mas_devops/roles/suite_app_config/README.md
```diff
@@ -26,6 +26,15 @@ Defines the app components and versions to configure in the application workspac
   For Health (standalone) the default is:
     `health=latest`
 
+### mas_app_optimizer_plan
```
durera, line 29: This should just be `mas_app_ws_plan` .. optimizer is the first app to support more than one install plan.  Let's word the documentation that this is a generic thing, the same way `mas_app_ws_spec` works, but is only used by manage at the moment.

alequint: I will update to be `mas_app_plan` as this is a Application level parameter 👍

alequint:
Tests done

image

ibm/mas_devops/playbooks/oneclick_add_mso.yml
```diff
@@ -0,0 +1,21 @@
+---
```
durera, line 1: No objection to adding this, but if we are delivering the playbook the doc for the playbook must also be created explaining where this fits into the bigger picture (e.g. what other oneclick playbooks must be ran first)

ibm/mas_devops/playbooks/oneclick_add_optimizer.yml
```diff
@@ -0,0 +1,22 @@
+---
```
durera, line 1: No objection to adding this, but if we are delivering the playbook the doc for the playbook must also be created explaining where this fits into the bigger picture (e.g. what other oneclick playbooks must be ran first)

ibm/mas_devops/roles/suite_app_config/vars/defaultspecs/optimizer.yml
```diff
@@ -0,0 +1,5 @@
+---
+# Default application spec for Optimizer
+mas_app_ws_spec:
+  bindings:
+    manage: workspace
```
durera, line 5: Optimizer shouldn't have this binding ... it's not valid to specify app to app bindings here, see dioscussion in Slack from last week where I believe it was confirmed this is not even needed.

ibm/mas_devops/roles/suite_app_install/README.md
```diff
@@ -16,6 +16,15 @@ Role Variables
 - `mas_app_id` Defines the kind of application that is intended for installation such as `assist`, `health`, `iot`, `manage`, `monitor`, `mso`, `predict`, or `safety`
 - `mas_app_upgrade_strategy` Defines the Upgrade strategy for the MAS Application Operator. Default is set to Automatic
 
+### mas_app_plan
+Defines what plan will be used in application install.
+
+- Environment Variable: `MAS_APP_PLAN`
+- Default: `full`
```
durera, line 23: Default can't be `full` ... for most apps the value here is irrelevent.

implementation (and doc) should be something like

> ### mas_app_plan
> Defines what plan will be used in application install.
> 
> - Optional
> - Environment Variable: `MAS_APP_PLAN`
> - Default: Application-specific, see details below.
> - Application Support:
>     - Optimizer v8.2+: `full` and `limited` are supported, defaults to `full`

ibm/mas_devops/roles/suite_app_install/defaults/main.yml
```diff
@@ -19,3 +19,4 @@ mas_entitlement_key: "{{ lookup('env', 'MAS_ENTITLEMENT_KEY') }}"
 
 # MAS application configuration
 mas_app_id: "{{ lookup('env', 'MAS_APP_ID') }}"
+mas_app_plan: "{{ lookup('env', 'MAS_APP_PLAN') | default('full', true) }}"
```
durera, line 22: We shouldn't default this here ... there's no sensible default for all apps.

durera (changes requested):

ibm/mas_devops/playbooks/oneclick_add_optimizer.yml
alequint: Agree. Added.

ibm/mas_devops/roles/suite_app_install/README.md
alequint: @durera maybe what you added above is just an example, but Optimizer will start in 8.2 following MSO 8.1 or will start from 8.0? @lee-cotton @yyjia-cn ?

ibm/mas_devops/roles/suite_app_install/defaults/main.yml
alequint: Agree. Removed.

ibm/mas_devops/roles/suite_app_config/vars/defaultspecs/optimizer.yml
alequint: Waiting for @lee-cotton 's double confirmation here: https://ibm-watson-iot.slack.com/archives/C02AF89FYF9/p1653427785449579?thread_ts=1653413356.140309&cid=C02AF89FYF9 (confusion between Manage and Workspace binding, not sure if it was meant to be the same behind the scenes)

alequint: Manage code is not already ready to remove it: https://ibm-watson-iot.slack.com/archives/C02AF89FYF9/p1653461536070859?thread_ts=1653413356.140309&cid=C02AF89FYF9

durera (dismissed):

ibm/mas_devops/roles/suite_app_config/vars/defaultspecs/optimizer.yml
```diff
@@ -0,0 +1,3 @@
+---
+# Default application spec for Optimizer
+mas_app_ws_spec:
```
durera, line 3: This is not valid  you only have half of the assignment.  need to assign empty object to this: `mas_app_ws_spec: {}`

durera (changes requested):

ibm/mas_devops/roles/suite_app_install/templates/operator-group.yml.j2
```diff
@@ -1,7 +1,7 @@
 apiVersion: operators.coreos.com/v1
 kind: OperatorGroup
 metadata:
-  name: {{ mas_app_namespace }}-operator-group
+  name: ibm-{{ mas_app_namespace }}-operatorgroup
```
durera, line 4: To avoid this being a breaking change in the collection, the code needs to be able to work with the old naming convention too.

e.g we know there can be only one operatorgroup in the namespace ... do a lookup for an operatorgroup, and if we find one, just use that name.  But if we don't use this name.  This way the collection will still work if you use it to update a previously installed version of MAS using the collection.  CoreAPI should behave the same way too as well, we shouldn't be so fragile to the name of the operatorgroup.

durera (changes requested):

ibm/mas_devops/roles/suite_app_install/templates/operator-group.yml.j2
alequint: @durera I have moved this change to the draft PR https://github.com/ibm-mas/ansible-devops/pull/317 so we can go ahead and merge this one sooner to bring back daily fvts as expected by João and team.

durera (approved):

Issue #288, opened 2022-05-02
Installation of health-aio should be configurable 


racree:
With the recent changes when MAS and Manage are installed Health-AIO is automatically installed.  Health-AIO requires MSO to be installed or it fails.  Therefore MAS and Manage cannot be installed without also installing Health-AIO and MSO.  
Whether or not to install Health-AIO needs to be a configuration option.

aroute:
This job fails out of the box with AIO enabled: `health-aio-`

durera:
Issue addressed in #316

PR #316, opened 2022-05-25
[patch] add aioFLAG

lokeshs09:
Set Default value of AIO to false and make it configurable via env. https://github.com/ibm-mas/ansible-devops/issues/288

### DIFF
--- a/ibm/mas_devops/playbooks/oneclick_add_manage.yml
+++ b/ibm/mas_devops/playbooks/oneclick_add_manage.yml
@@ -16,7 +16,7 @@
     db2_namespace: db2u
 
     # Application Installation
-    mas_app_id: manage
+    mas_app_id: "{{ lookup('env', 'MAS_APP_ID') }}"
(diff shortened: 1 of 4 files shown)


ibm/mas_devops/playbooks/oneclick_add_manage.yml
```diff
@@ -16,7 +16,7 @@
     db2_namespace: db2u
 
     # Application Installation
-    mas_app_id: manage
+    mas_app_id: "{{ lookup('env', 'MAS_APP_ID') }}"
```
durera, line 19: Can we default this to `manage` when not set please, so it's not a breaking change.

durera (changes requested):

ibm/mas_devops/playbooks/oneclick_add_manage.yml
lokeshs09: sure. done.

durera (approved):

PR #317, opened 2022-05-26
[patch] Rename Operator Group created by suite-app-install

alequint:
- [x] Rename OG
- [ ] Backward compatibility: use OG already available against the namespace in use... when none is found, use standard name


durera:
Closing PR due to old age/lack of activity

PR #323, opened 2022-06-01
[minor] Consolidate entitlement management

durera:
## Entitlement handling cleanup
Most users won't care about `SLS_ENTITLEMENT_KEY` vs `DB2_ENTITLEMENT_KEY` vs `MAS_ENTITLEMENT_KEY` etc ... they just want to provide `IBM_ENTITLEMENT_KEY`.

This update makes `IBM_ENTITELEMENT_KEY` a required property across the various roles that use require IBM entitlement in the install pipeline today:
- `sls`
- `db2`
- `suite_install`
- `suite_app_install`

The additional properties are still supported, but are now optional overrides for the default `ibm_entitlement_key` which will really only see use internally for development purposes.

## CP4D fixes
- When we set `spec.version` we also have to set `spec.ccs_operand_version` and `spec.datarefinery_operand_version` otherwise the inconsistency between the operand version we directly control and the operand version of the CP4D internals create incompatabilities (because CP4D patch versions can contain breaking changes).  With this update we can directly control the version of Watson Studio, CCS, and the Data Refinery.
- Missing `cpd_product_version` has been put into the Tekton tasks
- Standalone CP4D playbook has been added to the collection `ibm.mas_devops.cp4d`
- `ibm-entitlement-key` secret is installed in the CP4D instance namespace and the CP4D service accounts are patched to use it.  This removes the need to install a cluster-wide entitlement secret ([CPD-Quality#3478](https://github.ibm.com/PrivateCloud-analytics/CPD-Quality/issues/3478)).
- Watson Discovery reports a false install failure, updated automation to not exit early when Failure status is reported ([CPD-Quality#3480](https://github.ibm.com/PrivateCloud-analytics/CPD-Quality/issues/3480))

## Makefile
Added a simple makefile support the docker image and ansible collection to make it easier to do local development.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,46 +4,19 @@ Note that during a build the version of the ansible collection is automatically 
 
 
 ## Building the collection locally
+- Build and install the collection `make ansible-all` (default action)
+- Build the collection `make ansible-build`
+- Install the already built collection `make ansible-install`
 
+### Testing a role
 ```bash
-# Build
-ansible-galaxy collection build --output-path image/ansible-devops/ ibm/mas_devops --force && mv image/ansible-devops/ibm-mas_devops-11.0.0.tar.gz image/ansible-devops/ibm-mas_devops.tar.gz
-
-# Install
-ansible-galaxy collection install image/ansible-devops/ibm-mas_devops.tar.gz --force
-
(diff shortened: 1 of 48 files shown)


whitfiea (dismissed):
Looks good to me

PR #330, opened 2022-06-08
[minor] Support creation of Db2 LDAP user

Yolandabjyf:
This update allows the user to set up an additional Db2 user account by setting `db2_ldap_username` and `dp2_ldap_password` when creating a Db2 instance with the `db2` role.  These new properties are optional.


ibm/mas_devops/roles/db2/README.md
```diff
@@ -181,6 +181,17 @@ Define the Kubernetes memory limit for the Db2 pod.
 - Environment Variable: `DB2_MEMORY_LIMITS`
 - Default: `12Gi`
 
+### [optional] db2_ldap_username
```
durera, line 184: Please follow the same convention used in the other variables in the Readme ... e.g for the first one it should be:

```
### db2_ldap_username
Define the username of db2 in the local LDAP registry. If this defined, LDAP user will be the Manage user used to configure the JDBC configuration on MAS.

- Optional
- Environment Variable: `DB2_LDAP_USERNAME`
- Default: None
```

durera (changes requested):

durera (approved):

Issue #340, opened 2022-06-11
Support Cloudflare DNS integration


durera:
As an alternative to IBMCloud CIS, this will likely involve some rejig of how the `suite_dns` role works to present a more generic interface than the one that only factored in CIS.

This integration should include registration of the DNS ANAME records, so that no manual actions are required.

Useful links
- https://api.cloudflare.com/
- https://cert-manager.io/docs/configuration/acme/dns01/#setting-nameservers-for-dns01-self-check

ClusterIssuer will look something like:

```
apiVersion: cert-manager.io/v1
kind: ClusterIssuer
metadata:
  name: letsencrypt-prod
spec:
  acme:
    # The ACME server URL
    server: https://acme-v02.api.letsencrypt.org/directory
    email: xxx
    privateKeySecretRef:
      name: letsencrypt-prod
    solvers:
    - dns01:
        cloudflare:
          email: xxx
          apiTokenSecretRef:
            name: cloudflare-api-token-secret
            key: api-token
```

With the `cloudflare-api-token-secret`:
```
apiVersion: v1
kind: Secret
metadata:
  name: cloudflare-api-token-secret
type: Opaque
stringData:
  api-token: xxxxxx
```

belfortmaycon:
Please, add to Digital Ocean too. It is the same as Cloudflare procedure

PR #349, opened 2022-06-14
[minor] Add support for Cloudflare DNS integration

durera:
Updates the `suite_dns` role to add Cloudflare as a supported DNS provider, also cleans up the implementation of the IBM Cloud Internet Services (CIS) provider to align with the v10 design.  Both providers are supported by the Tekton tasks as well now.

- #340


jjiraud:
How do we get the Open shift ingress subdomain(external route) to set up the cname record in cloudflare. I cannot see the or get the subdomain in OCP provisioned on AWS.

Regards,
Jean

PR #377, opened 2022-06-29
[patch] Remove extra prefix in project name

alequint:


alequint:
I merging that to take advantage of current test in progress of pre-master, so we can update image before suite-app-config is pulled

PR #404, opened 2022-07-14
[patch] fix db2_backup role (var name mismatch)

cuddlyporcupine:
- Bug fix: mismatch variable name for db2 pod
- Bug fix: backup was written to ephemeral storage inside db2 pod, resulting in pod restart for large database backups.  backup is now written to the backup pvc of the db2 pod


ibm/mas_devops/roles/db2_backup/defaults/main.yml
```diff
@@ -2,6 +2,7 @@
 # Db2 backup default vars
 db2_backup_dir: "{{ lookup('env', 'DB2_BACKUP_DIR') }}"
 db2_backup_instance_name: "{{ lookup('env', 'DB2_BACKUP_INSTANCE_NAME') }}"
+db2_namespace: "{{ lookup('env', 'DB2_NAMESPACE') }}"
```
alequint, line 5: @cuddlyporcupine  To be consistent with default values used in db2 role, I would suggest we default namespace in backup/restore roles with same value as [here](https://github.com/ibm-mas/ansible-devops/blob/master/ibm/mas_devops/roles/db2/defaults/main.yml#L4)

ibm/mas_devops/roles/db2_backup/README.md
```diff
@@ -33,6 +39,8 @@ Example Playbook
 - hosts: localhost
   any_errors_fatal: true
   vars:
+    db2_namespace: "db2u"
```
alequint, line 42: @cuddlyporcupine  To be consistent with default values used in db2 role, I would suggest we default namespace in backup/restore roles with same value as [here](https://github.com/ibm-mas/ansible-devops/blob/master/ibm/mas_devops/roles/db2/defaults/main.yml#L4)

alequint (changes requested):

ibm/mas_devops/roles/db2_backup/defaults/main.yml
cuddlyporcupine: @alequint requested change for default is in

alequint (approved):

durera:
Also, can you write up the summary description for this PR please .. doesn't need an essay, just a paragraph (or even a sentence) sumarising what is being changed and why.

alequint (dismissed):

alequint:
@cuddlyporcupine dicker-build action is failing, but not sure why... happening only in this PR/branch. Can you check?

PR #420, opened 2022-07-28
[minor] Support Manage attachments and BIM configuration

andrercm:
This PR introduces changes needed to reduce the manual steps needed while setting up IVT environments and more specifically around Manage deployment:

- Adds support to configure COS instance to be used as storage location for Manage application attachments (This is something we need out of the box in all Manage IVT environments)
- Adds support to define base and secondary languages for Manage application (Also something needed OOTB for Manage IVT environments)
- Adds support to define different flavors of server bundle configurations for Manage application (Manage requires this to be flexible accordingly to the expected workload)
- ~Changes the default scope for Manage JDBC config to be `workspace-application` instead of `system` to make it consistent with what users get while deploying Manage from MAS UI.~
- Added also support for BIM configuration using default File Storage PVC mount path.


ibm/mas_devops/roles/suite_app_config/README.md
```diff
@@ -75,9 +75,44 @@ Optional. Name of the schema where Manage database lives in.
 ### mas_app_settings_demodata
 Optional. Flag indicating if manage demodata should be loaded or not.
 
-- Environment Variable: MAS_APP_SETTINGS_DEMODATA
+- Environment Variable: `MAS_APP_SETTINGS_DEMODATA`
 - Default: `false` (do not load demodata)
 
+### mas_app_base_language
```
alequint, line 81: Just to keep consistent with other variables also used in application settings, can we use prefix `mas_app_settings_` instead of just `mas_app_` ? Same for the next two variables:

- `mas_app_settings_base_language`
- `mas_app_settings_secondary_languages`
- `mas_app_settings_server_bundles_size`

ibm/mas_devops/roles/suite_app_config/README.md
```diff
@@ -97,7 +132,7 @@ Example Playbook
 
     mas_app_ws_spec:
       bindings:
-        jdbc: "{{ mas_appws_jdbc_binding | default( 'system' , true) }}"
+        jdbc: "{{ mas_appws_jdbc_binding | default( 'workspace-application' , true) }}"
```
alequint, line 135: Why change the default? Can we have system (the most embracing scope) and set workspace-application as an exception case via environment variable? Any particular reason to use `workspace-application` scope and not `system`?

alequint (changes requested):

ibm/mas_devops/roles/cos_attachments_config/meta/main.yml
```diff
@@ -0,0 +1,21 @@
...
+      versions:
+        - all
+
+  galaxy_tags:
+    - ibm
+    - mas
+    - devops
+    - rhocp
+
+dependencies: []
```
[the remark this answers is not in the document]
andrercm: fixed

ibm/mas_devops/roles/suite_app_config/README.md
andrercm: Well that's just because Manage's database is configured by default at workspace level for Manage (you setup Manage database not during Manage operator deployment page, but during Manage activation page...), so just thought about having it consistent with what the playbooks should do? I don't think there's any different tbh, If you see any problems with that I can remove it and for IVT envs i just set this via env var... i mean the ansible var with `mas_appws` prefix is really something that correlates to `workspace` right? So this is really just about consistency.

ibm/mas_devops/roles/suite_app_config/README.md
andrercm: fixed

andrercm: fixed!

ibm/mas_devops/roles/suite_app_config/README.md
durera: Keep default as system .. we don't want to worry about the knock on effects of changing this really.  It just complicates getting this merged in.  In general the default mode in this collection is "deploy for the smallest footprint", which would mean sharing databases.

alequint (approved):

alequint:
@andrercm super! 👏

I see you have also configured BIM pvc and already mounted the volumes as needed by ManageWorkspace CR right?

I did not test in my environment but I am sure you did. Everything looks good to me now.

durera (approved):

ibm/mas_devops/roles/suite_app_config/vars/defaultspecs/manage.yml
```diff
@@ -5,6 +5,28 @@ mas_app_ws_spec:
     jdbc: "{{ mas_appws_jdbc_binding | default( 'system' , true) }}"
   components: "{{ mas_appws_components | default({'base':{'version':'latest'}}, true) }}"
   settings:
+    deployment:
```
durera, line 8: Bit concerned about this?  Shouldn't this new block be templated behind a {% if condition %}, otherwise this becomes the default for everyone, not just those who are doing this new stuff?

durera: Are all the defaults if I do nothing differently to before going to result in a compatible Manage install to what I had before this change?

andrercm:
@durera @alequint anything missing so we can merge this?

durera (approved):

PR #421, opened 2022-07-29
[minor] Add Single Node OpenShift on AWS support

bzzwh:
created playbooks for aws sno provision and deprovision


durera (commented):
I'll give this a test drive this week, and we'll merge it in as part of the v11 release if all goes well :+1:

PR #457, opened 2022-08-30
[patch] Support OCP_INGRESS_TLS_SECRET_NAME

racree:
Add support for optional `OCP_INGRESS_TLS_SECRET_NAME` environment variable which will allow the user to explcitly control the name of the secret containing the OCP default ingress TLS certificate.  This is only required when the secret name does not match any of the known patterns that are supported out of the box by the collection.

### DIFF
--- a/ibm/mas_devops/common_tasks/get_ingress_cert.yml
+++ b/ibm/mas_devops/common_tasks/get_ingress_cert.yml
@@ -1,12 +1,16 @@
 ---
 # 1. Lookup for router-certs-default secret
 # -----------------------------------------------------------------------------
+- name: set ocp ingress tls secret
+  set_fact:
+    ocp_ingress_tls_secret: "{{ lookup('env', 'OCP_INGRESS_TLS_SECRET') | default('router-certs-default', True) }}"
+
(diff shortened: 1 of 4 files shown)


ibm/mas_devops/common_tasks/get_ingress_cert.yml
```diff
@@ -1,12 +1,16 @@
 ---
 # 1. Lookup for router-certs-default secret
 # -----------------------------------------------------------------------------
+- name: set ocp ingress tls secret
+  set_fact:
```
alequint, line 5: @racree thinking again here... I would add a `when` clause here to only to this set_fact when `ocp_ingress_tls_secret` is not defined. I am thinking that this var must be set in a playbook or in a role calling such common task (see documentation comment within this same review)...

alequint (dismissed):
@racree thinking on the docs, as you updated common_tasks and given that we do not have a documentation specifically for that, I would add the new `OCP_INGRESS_TLS_SECRET` as part of the README.md of all the three roles that use this common task as an optional variable to be used.

@durera pls dbl check this review

durera (approved):

Issue #461, opened 2022-09-12
Modify nvidia_gpu role to support OCP 4.10


alequint:
As per Dominique Clain:

```
This role is not valid anymore for OCP 4.10 (only for OCP 4.6/4.8), we'll need to update the channel and cluster policy.
```

durera:
This was already done, but not linked back to the github issue here.

PR #463, opened 2022-09-12
[patch] Update gpu cluster policy for OCP 4.10

dclain:
https://github.com/ibm-mas/ansible-devops/issues/461


ibm/mas_devops/roles/nvidia_gpu/templates/clusterpolicy.yml.j2
```diff
@@ -4,153 +4,66 @@ kind: ClusterPolicy
 metadata:
   name: gpu-cluster-policy
 spec:
-  daemonsets:
```
alequint, line 7: @dclain I see you are updating role to worth with OCP 4.10... question: don't we need to support 4.8 as well as MAS still support it? Because if so, I presume the policies, versions and namespaces used before must still be supported right? Unless MVI claim support only for OCP 4.10?

ibm/mas_devops/roles/nvidia_gpu/tasks/main.yml
```diff
@@ -26,7 +26,7 @@
 # -----------------------------------------------------------------------------
 - name: "Create gpu operator group"
   when:
-    - gpu_namespace != "openshift-operators"
```
alequint, line 29: Same question as in cluster policy comment.

ibm/mas_devops/roles/nvidia_gpu/templates/clusterpolicy.yml.j2
dclain: yeah I think we possibly need templates per OCP version I guess... cause the cluster policy will also changes for OCP 4.11. Any existing roles we can piggy back on that shows different roles/templates per OCP version?

durera: @dclain I'd rather we looked it up, but it seems the other roles that rely on this took the easy option:

```
# We need to know what the OCP_RELEASE is to know what channel to target the subscriptions at
ocp_release: "{{ lookup('env', 'OCP_RELEASE') | default('4.10', True) }}"
```

I suggest for now do the same, and when this is hopefully improved in the future it can be improved across the collection as a whole.

alequint:
@caroazad could you help to complete @dclain changes here so we can close this PR? We need to do the necessary updates for backward compatibility of nvidua gpu operator. That is the missing part to merge.

alequint (changes requested):
@caroazad could you help to complete @dclain changes here so we can close this PR? We need to do the necessary updates for backward compatibility of nvidua gpu operator. That is the missing part to merge.

dclain:
I'm closing this PR as Caroline is addressing those changes thru PR https://github.com/ibm-mas/ansible-devops/pull/525

PR #466, opened 2022-09-14
[patch] - Support CloudPak for Data 4.5.x installs/upgrade

andrercm:
- Introduces support for CPD 4.5.x installs, therefore with this change both CPD 4.0.9 and CPD 4.5.x can be installed via our collection by simply specifying `cpd_version: 4.0.9` or `cpd_product_version:4.0.9` (to install CPD control plane + CPD services at the same version)
- Also adds support to upgrade path (CPD 4.0.x -> 4.5.x), currently supported versions are 4.0.9, 4.5.0, 4.5.1 and 4.5.2


ibm/mas_devops/roles/cp4d/README.md
```diff
@@ -1,13 +1,34 @@
 cp4d
 ====
 
-This role installs [IBM Cloud Pak for Data](https://www.ibm.com/uk-en/products/cloud-pak-for-data) Operator in the target cluster.
+This role installs or upgrades [IBM Cloud Pak for Data](https://www.ibm.com/uk-en/products/cloud-pak-for-data) Operator in the target cluster.
 
-The [cp4d_hack_worker_nodes](cp4d_hack_worker_nodes.md) role must have been executed during cluster set up to update the cluster's global image pull secret and reload all worker nodes.  Unfortunately Cloud Pak for data does not support using image pull secrets attached to service accounts in the namespace.
+Currently supported Cloud Pak for Data release versions supported are:
```
alequint, line 6: minor: eliminate one of the "supported"  words

ibm/mas_devops/roles/cp4d_service/defaults/main.yml
```diff
@@ -8,7 +8,9 @@
 # of our control -- the only version of CP4D that works with MAS 8.7 is v4.0.7, but we can't install that
 # from the IBM operator catalog at present, much more work is needed in this collection to support the operating
 # model of the Cloud Pak for Data team :/
-cpd_product_version: "{{ lookup('env', 'CPD_PRODUCT_VERSION') }}"
+cpd_product_version: "{{ lookup('env', 'CPD_PRODUCT_VERSION') | default(lookup('env', 'CPD_RELEASE_VERSION'), true) | default('4.0.9', true)  }}"
```
alequint, line 11: Will we keep CP4D 4.5 as an exception after merge this PR and eventually change the default to it I presume?

alequint:
@andrercm not much to say for this review... I think what we really can do next before we merge this is run a complete pipeline using your branch code and setting CPD_RELEASE_VERSION to 4.5 and see what we get... that way we anticipate what Assist, Predict and HPU activation/configuration will find out. Passing that stage then we need to roll these change over MAS CLI and Ansible-FVT

ibm/mas_devops/roles/cp4d/README.md
andrercm: fixed!

ibm/mas_devops/roles/cp4d_service/defaults/main.yml
andrercm: That's what I believe we'll do... probably when MAS 8.9 is released, CPD 4.5 will become the default.
@durera is that a fair assumption?

alequint, line 11: I vote for simplifying variables here. Suggest to use cpd_release_version / CPD_RELEASE_VERSION in both cpd and cpd_service roles. We don't want users setting conflicting values here, it does not make sense for what we need in MAS.

@durera any comments here?

ibm/mas_devops/roles/cp4d/defaults/main.yml
```diff
@@ -31,3 +31,14 @@ cpd_entitlement_key: "{{ lookup('env', 'CPD_ENTITLEMENT_KEY') | default(ibm_enti
 # for the Cloud Pak for Data operators.
 cpd_operators_namespace: "{{ lookup('env', 'CPD_OPERATORS_NAMESPACE') | default('ibm-cpd-operators', true) }}"
 cpd_instance_namespace: "{{ lookup('env', 'CPD_INSTANCE_NAMESPACE') | default('ibm-cpd', true) }}"
+cpd_release_version: "{{ lookup('env', 'CPD_RELEASE_VERSION') | default(lookup('env', 'CPD_PRODUCT_VERSION'), true) | default('4.0.9', true)  }}"
```
alequint, line 34: I vote for simplifying variables here. Suggest to use cpd_release_version / CPD_RELEASE_VERSION in both cpd and cpd_service roles. We don't want users setting conflicting values here, it does not make sense for what we need in MAS.

@durera any comments here?

alequint (changes requested):

ibm/mas_devops/roles/cp4d/defaults/main.yml
andrercm: removed all the `cpd_release_version` references and just kept the `cpd_product_version` thru the CP4D roles.

alequint (approved):

PR #474, opened 2022-09-23
[patch] Support for AppConnect upgrades

andrercm:
- Support for AppConnect seamlessly upgrades.
- Introduces default AppConnect settings based on MAS HPU versions.


alequint:
@andrercm @durera  let's hold this one till we investigate compatibility issues between AppConnect versions and specific OCP versions. So far, v. 5.2 cannot install in OCP 4.8 and OCP 4.9.... test using a clean OCP 4.8 is in progress (cluster `appconnect-test` in MAS Dev Account)

andrercm:
@alequint during my tests v5.2 only worked when we upgraded to OCP 4.10

PR #475, opened 2022-09-28
Fix broken docker run command

vincent-pli:
The command not work as expected


alequint (approved):

alequint:
Thanks for the correction @vincent-pli 

Reference: https://docs.docker.com/engine/reference/run/#clean-up---rm

Approved and merge it

PR #477, opened 2022-10-03
[patch] - Support uninstall resources

andrercm:
Introduces two generic roles to enable us to uninstall custom resources and uninstall subscription and its csv's.


durera (changes requested):
I suggest moving `uninstall_cr` and `uninstall_task` to the `common_tasks` directory e.g. see `common_tasks/detect_airgap.yml` .. so that we do not need to expose this as external roles for a user, but rather they are re-usable internals for use in other roles.

durera:
Superceeded by work in https://github.com/ibm-mas/ansible-devops/tree/uninstallv2

PR #479, opened 2022-10-14
[minor] Allow DB2U Cluster settings to be customized

andrercm:
This fix is to address the problem identified with the defaulted `INSTANCE_MEMORY` being insufficient to support DB2 load for Manage/IoT databases provisioned by ansible devops collection. Also this allows the db2wh route to be created based on db2 instance name, so allows us to create multiple db2wh and corresponding routes for it.


alequint:
Setting PR to draft as it should not be merged

andrercm:
Eric's comments:
```
i was looking at https://github.com/ibm-mas/ansible-devops/pull/479 and wondering if it would be possible to allow the user of our db2 role to override the sections listed below in the CR.  This way it gives them more flexibility to configure various config items at these 3 different levels.
spec.environment.database.dbConfig
spec.environment.instance.dbmConfig
spec.environment.instance.registry
```

andrercm:
@cuddlyporcupine @alequint @durera this PR is also ready for review now. Been noticing db2 instances crashing due lack of heap memory cross IVT environment. This PR should take care of this problem by changing the current db2 heap size by default and also allow the user of our db2 role to override the sections listed below in the CR.  This way it gives them more flexibility to configure various config items at these 3 different levels.
```
spec.environment.database.dbConfig
spec.environment.instance.dbmConfig
spec.environment.instance.registry
```

ibm/mas_devops/roles/db2/templates/tlsroute.yml.j2
```diff
@@ -1,7 +1,7 @@
 kind: Route
 apiVersion: route.openshift.io/v1
 metadata:
-  name: db2u-tls-route
+  name: db2u-{{ db2_instance_name }}-tls-route
```
durera, line 4: We need to put cleanup logic into the role to remove the old route, otherwise when running an over-the-top install, user will be left with two routes that conflict.

- If possible we delete the old route after creating the new one, **but we need to test whether the two routes can co-exist**  If they cannot co-exist we need to delete the old route first, and accept a brief outage.

durera (changes requested):

ibm/mas_devops/roles/db2/templates/tlsroute.yml.j2
andrercm: I think it's okay to have two routes pointing to the same service, but I'll test and get back to you.

andrercm: Glad you mentioned this... indeed i get an issue with code `HostAlreadyClaimed ` if having two routes exposing the same endpoint location.

image

Good thing is that in a scenario with two routes pointing to same host/endpoint, while existing route keeps active and serving the requests, the new created one conflicts and goes into error state. So, deleting old route replaces the endpoint automatically to the newly created route, with no downtime.

image

Therefore, as you suggested, i implemented the clean up step to remove old hardcoded `db2u-tls-route` after creating the new custom route, and only if the host/endpoint conflicts between old and new routes. That way we ensure that we only clean up the route in a conflicting host scenario.

durera (dismissed):

PR #483, opened 2022-10-19
[minor] Add support for OCP IPI on GCP

vskhabani:
Added changes related to GCP IPI support in ocp_provision and ocp_deprovision role.

OpenShift Container Platform requires a Google Cloud Platform (GCP) service account that provides authentication and authorization to access data in the Google APIs. Service Account file will be a json file which can be downloaded from GCP console.
By default openshift-installer checks service account into the env variable `GOOGLE_APPLICATION_CREDENTIALS` .
`gcp_service_account_file`  variable takes value present in the env `GOOGLE_APPLICATION_CREDENTIALS`

In ipi install-config.yaml.j2 added condition to use project_id if ipi_platform is gcp which is required specifically for the GCP.

Specific Openshift documentation: https://docs.openshift.com/container-platform/4.7/installing/installing_gcp/installing-gcp-customizations.html

Used this role to provision OpenShift Cluster on GCP account.


whitfiea (commented):
@vskhabani why is the env var called `GOOGLE_APPLICATION_CREDENTIALS` and not `GCP_SERVICE_ACCOUNT_FILE`? How is that file then used as the only changes seem to be to check the file exists but doesn't get used anywhere?

whitfiea (commented):
@vskhabani so I did some googling the it seems that GOOGLE_APPLICATION_CREDENTIALS is the expected env var in a GCP environment, and we are just assigning it to `gcp_service_account_file` to do a check only in our ansible?

whitfiea (approved):

durera (changes requested):
Please add an update to the changelogs for 11.5 (which this version will be), and this is good to merge:

here: https://github.com/ibm-mas/ansible-devops/tree/ipi-gcp/ibm/mas_devops#change-log
and here: https://github.com/ibm-mas/ansible-devops/blob/ipi-gcp/docs/changes.md#changes

vskhabani:
@durera Updated the changelogs for 11.5

docs/changes.md
```diff
@@ -1,6 +1,8 @@
 ## Changes
 Note that links to pull requests prior to public release of the code (4.0) direct to IBM GitHub Enterprise, and will only be accessible to IBM employees.
-
+- `11.5` Multiple Updates:
```
durera, line 3: Don't need to seperate out provision and deprovision, one entry per PR basically .... should just be:
```
- `11.5` Add support for OCP IPI on GCP ([#483](https://github.com/ibm-mas/ansible-devops/pull/483))
```

durera (changes requested):

docs/changes.md
vskhabani: Updated.

PR #486, opened 2022-10-20
[minor] Support default JMS queue config in Manage bundle workload config

andrercm:
This change adds support for additional server configuration for Manage bundle deployment which is required to setup Manage JMS queues.

Besides of enabling Manage to be deployed using multiple bundle pods including JMS dedicated workload, we need to add specific default configuration to setup Manage's jms messaging queues. The work consists on applying and assigning default secrets containing server config xml to the Manage bundle pods and JMS bundle pod specifically.

This will also allow us to have JMS configuration set by default whenever we want to deploy Manage along with our ansible collection.


ibm/mas_devops/playbooks/oneclick_add_assist.yml
```diff
@@ -15,7 +15,7 @@
 
     # Application Installation
     mas_app_id: assist
-    mas_app_channel: "{{ lookup('env', 'MAS_APP_CHANNEL') | default('8.5.x', true) }}"
+    mas_app_channel: 8.4.x
```
durera, line 18: Bad merge I think?

durera (changes requested):
This should be a minor .. you are adding functionality ... please update the changelogs to add an entry for this PR and make the change in oneclick_add_assist.yml that looks like a bad merge

ibm/mas_devops/playbooks/oneclick_add_assist.yml
andrercm: yes, fixed it

PR #492, opened 2022-10-26
[patch] Tweak Manage's attachment configuration

andrercm:
As per Manage team's (Clarissa) request, this PR unsets by default the property `mxe.report.AttachDoc.validateURL` in Maximo attachments db config.

We already set Manage config to secure attachments, so for Manage it makes sense to set it to `mxe.report.AttachDoc.validateURL = 0` along with it.


andrercm:
@alequint @durera can we merge this PR? this is just a minor setting that Manage teams asked me to include by default as we're are using Manage's secure attachment as `true` by default so `validateURL` should set set to `0` as well.

```
db2 "UPDATE MAXIMO.MAXPROPVALUE SET PROPVALUE = 'true' WHERE PROPNAME = 'mxe.doclink.securedAttachment';"
db2 "UPDATE MAXIMO.MAXPROPVALUE SET PROPVALUE = '0' WHERE PROPNAME = 'mxe.report.AttachDoc.validateURL';"
```

durera (approved):

PR #498, opened 2022-10-27
[patch] wait for WatsonDiscovery

racree:
Adding sleep for 5 minutes after CRDs are defined.


alequint (approved):

alequint:
Investigating a different issue that happen when installing Discovery; depending on outcome we may have more to come in this PR. Converting PR to DRAFT in the mean time

Issue #500, opened 2022-11-03
Implement support for manual cert management in ansible-devops and One-click installer 


sanju7216:
Adding support to set flag to enable/disable manual cert management while MAS installation and allow passing tls certs yaml 

so that it can be used for 2 reasons.
- [x] customers can actually install MAS using the new installer with manual certs instead of cert-manager
- [x] we can have FVT use manual certs (**Update**: Squad CI working on it. ETA May 5th)

PR #501, opened 2022-11-04
[minor] Add support for manual public cert management

sanju7216:
Adding support to enable manual public certificate management mode across the suite.
Using manual cert mgmt mode allows users to use their own public tls certificates rather than using IBM Cert manager's.

Manual mode can be enabled by adding this env var `MAS_MANUAL_CERT_MGMT=True` 
Also, the certs need to be added 
```
MAS_CONFIG_DIR/certs/<app>/tls.crt
MAS_CONFIG_DIR/certs/<app>/tls.key
MAS_CONFIG_DIR/certs/<app>/ca.crt
```

A new role `suite_certs` is added to handle the TLS cert secret creation in core and all other applications' namespace. This secret will then be processed by their respective operators.

`install_suite` role uses the new env var `MAS_MANUAL_CERT_MGMT` and adds a new field `spec.settings.manualCertMgmt` boolean field to suite CR to determine the certificate mgmt mode across the suite.

-----

It's been Included also in suite_certs role the possibility to use cis as dns provider, taking leverage of existing automation to manage all cnames and ingress redirection from it. It is optional, though, as user may have that already set manually when decided to use their own certificates in MAS.

Issue: https://github.com/ibm-mas/ansible-devops/issues/500

-----

- Tested in the pipeline

image

- Accesing the environment via the new route (Core, Manage and Assist)

image


durera (changes requested):
As discussed on Slack

ibm/mas_devops/roles/suite_install/templates/core_v1_suite.yml.j2
```diff
@@ -23,6 +23,7 @@ spec:
   license:
     accept: true
   settings:
+    manualCertMgmt: {{  mas_manual_cert_mgmt  }}
```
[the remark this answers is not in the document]
alequint: Already added by Sanjay here https://github.com/ibm-mas/ansible-devops/blob/manualcertmgmt/ibm/mas_devops/roles/suite_install/templates/core_v1_suite.yml.j2#L31

ibm/mas_devops/rebuild.sh
alequint, line 1: To be removed before creating official PR

ibm/mas_devops/roles/cert_manual_order/meta/main.yml
```diff
@@ -0,0 +1,22 @@
+galaxy_info:
```
alequint, line 1: This role may fit better in ansible-fvt as it is very specific for an initial load of CNAMES in CIS to be used by Secrets Manager to generate the certificate

ibm/mas_devops/roles/suite_certs/meta/main.yml
```diff
@@ -0,0 +1,21 @@
+galaxy_info:
+  author: Sanjay Prabhakar & Alexandre Quinteiro
```
durera, line 2: This should be converted to a list:

```
author: 
  - Sanjay Prabhakar
  - Alexandre Quinteiro
```

ibm/mas_devops/roles/suite_certs/tasks/main.yml
```diff
@@ -0,0 +1,10 @@
+---
+
+- name: "Debug mas_manual_cert_mgmt"
+  when: not mas_manual_cert_mgmt
+  debug:
+    msg: "mas_manual_cert_mgmt is not True, indicating the installation is using cert-managet to create certificates. This role will not take any action"
```
durera, line 6: Typo in message

durera (changes requested):
I'm not convinced about the relationship between the two roles `suite_certs` and `suite_dns` at this point, it feels a bit hokey  .. we might in the future want to look at combining them given the overlap we now have, but right now it's more important to get support for this feature integrated after the delays.

There's a few minor things to fix before we merge, see individual comments.

ibm/mas_devops/roles/suite_certs/meta/main.yml
durera: Fixed

ibm/mas_devops/roles/suite_certs/tasks/main.yml
durera: Fixed

durera (approved):

PR #505, opened 2022-11-08
[patch] Fix issue that install  conflicting versions

alequint:
I have noticed problems during installation of common service operator; cpd, uds and certmanager: all of them timed out during install pipeline. Then checking installed operators under ibm-common-services namespace, we see this:

image

We noticed that 1.17 subscribe to v3 channel while 1.18 is in v3.20... looking at the code, I believe the issue is exactly the version we set today in our subscription (v3 and not v3.20) - it seems that when v3 is subscribed it tries to install both versions - bizarre behavior.

- [ ] Test fix to be sure it solve the issue


durera (changes requested):
Can't just change this to 3.20 ... it will stop working for older catalogs that only have v3.

alequint:
@durera I don't know even if by changing to 3.20 will solve the problem... need to test. Important to say that in OCP 4.8 we have also v3.20 available; 4.6 is no longer supported, so we may need to check in 4.7; of course, IF changing the channel will solve the issue (but as we cannot reproduce it all the times, it gets complicated)... will keep an eye on the next runs anyway (I see it happening with CPD 4.5 today twice, but not sure if other times were 4.5 as well or 4.0.9; will pay attention on it)

alequint:
Closing, this is not an actual fix for this problem. Refer to https://ibm-watson-iot.slack.com/archives/C031Q7W21FZ/p1669682373886519?thread_ts=1669636196.814199&cid=C031Q7W21FZ

PR #512, opened 2022-11-11
[minor] Updates to support blocking Artifactory

racree:


ibm/mas_devops/roles/install_operator/templates/ibm-entitlement-with-artifactory.json.j2
```diff
@@ -2,4 +2,4 @@
 {% if artifactory_username is defined and artifactory_username != "" %}
 "wiotp-docker-local.artifactory.swg-devops.com":{"username":"{{ artifactory_username }}","password":"{{ artifactory_apikey }}","auth":"{{ (artifactory_username ~ ':' ~ artifactory_apikey) | b64encode }}"},
 {% endif %}
-"cp.icr.io/cp":{"username":"{{ icr_username }}","password":"{{ icr_password }}","auth":"{{ (icr_username ~ ':' ~ icr_password) | b64encode }}"}}}
+"{{ icr_host }}":{"username":"{{ icr_username }}","password":"{{ icr_password }}","auth":"{{ (icr_username ~ ':' ~ icr_password) | b64encode }}"}}}
```
durera, line 5: This was deliberately changed to be explicity ICR ... if we are using artifactory as the alternative source for ICR content then both credentials will be in the secret now .. this allows for mixed content (some from ICR and some from Artifactory) to work in the same namespace (e.g. using pre-release MAS code with GA truststore manager)

ibm/mas_devops/roles/suite_install/README.md
```diff
@@ -11,10 +11,10 @@ Role Variables
...
 - `mas_instance_id` Defines the instance id to be used for MAS installation
-- `mas_icr_cp` Defines the entitled registry from the images should be pulled from. Set this to `cp.icr.io/cp` when installing release version of MAS or `wiotp-docker-local.artifactory.swg-devops.com` for dev
-- `mas_icr_cpopen` Defines the registry for non entitled images, such as operators. Set this to `icr.io/cpopen` when installing release version of MAS or `wiotp-docker-local.artifactory.swg-devops.com` for dev
-- `mas_entitlement_username` Username for entitled registry. This username will be used to create the image pull secret. Set to `cp` when installing release or use your `w3Id` for dev.
-- `mas_entitlement_key` API Key for entitled registry. This password will be used to create the image pull secret. Set to with IBM entitlement key when installing release or use your artifactory `apikey` for dev.
+- `mas_icr_cp` Defines the entitled registry from the images should be pulled from. Set this to `cp.icr.io/cp` when installing release version of MAS or `wiotp-docker-local.artifactory.swg-devops.com` for dev or the local registry endpoint when installing in an air gap environment.
```
durera, line 14: The user does not need to set this in an airgap environment where image mirroring (and reference by digest) uses the ImageContentSourcePolicy to auto-redirect from the original source to the mirror; it's only used in development to set up Artifactory as an alternative for ICR.

durera (changes requested):

PR #520, opened 2022-11-16
[patch] Increase default Db2 resource requirements

andrercm:
We did an initial db2 resource tweak to prevent db2 instances provisioned as part of this collection to break due memory shortage, however that was not enough. After talking with @ctrafa, he still got memory issues while instantiating db2 using this role, so to fix it, we had to increase also the db2 resource limits that are used by default.


andrercm:
@ctrafa @durera this is also ready for review. This is a db2 tweak in resource limits to avoid db2 from breaking due memory shortage.

durera:
How are we not hitting issues with the current settings every day in daily FVT, or are we and no-one has noticed?  

Doubling the default requests is a significant change.  Given we have the ability to customize the requests, why do we need/want to change the defaults?  Is there some documentation for Db2 that recommends a minimum of 4CPU and 12Gb memory for Db2 justifying the much larger default setting now,or have we since proven that Db2 can't function propely on 2CPU and 6Gb memory so it's a bad default?  

I don't mind changing the default, but we should have a justification for why if we are changing this - because we are going to change the default for every user of this collection.

andrercm:
@durera honestly i don't have much reasons to sustain why we need to double it, i think we'd need some sort of DBA experts to help us assess. However my main concern is how easily we are finding DB2 instances hitting memory issues while installing IoT/Predict/Manage against it. It has happened recently in a customer environment that @ctrafa was helping to get fixed and over the course of the IVT cycle we also have experience this type of memory shortage across all shared envs. My guess of why we're not hitting this in FVT is due the fact that they are daily being recreated, that type of issue is mostly seen in envs running for some time.

durera (approved):

PR #523, opened 2022-11-18
[patch] Update digest maps before updating catalogsource

durera:
This update addressed a flaw discovered in the update process in a disconnected MAS install.

When a catalog source is updated in a connected install, that is the only action required.  The operators will update and will in turn update the operands under their control.

However, in a disconnected install the need for the additional configmap containing the digest map references must be updated before the operator is updated, otherwise when the operator updates it will fall back to using image tags instead of digests, and will be unable to resolve the container images from the mirror registry.

The need for this additional config map to be installed is in the process of being eliminated, but at present most MAS operators require this step.

This update integrates the action of updating the configmaps for all installed (supported) operators.


durera:
Fix validated

PR #528, opened 2022-11-29
[minor] Add support of SSH public key in ocp_provision role

vskhabani:
Added **sshKey** variable which will have value of a custom Public SSH key.
If the user gives public ssh key value then that will be added to the OCP cluster nodes (Master and workers).
Using the private key user can then SSH to the OCP cluster nodes using a bastion host for troubleshooting.

[documentation ](https://docs.openshift.com/container-platform/4.6/networking/accessing-hosts.html)


whitfiea (commented):
@vskhabani Do you need to update the docs to say what this setting is needed for etc? Plus for a minor it needs a changelog update

vskhabani:
Hi @whitfiea , I will change it to patch then. Also will update the readme

whitfiea (approved):

ibm/mas_devops/roles/ocp_provision/templates/ipi/install-config.yaml.j2
```diff
@@ -31,3 +31,7 @@ platform:
 {% endif %}
 publish: External
 pullSecret: '{{ lookup('file', ipi_pull_secret_file) | regex_replace('\'', '"') }}'
+{% if sshKey != "" %}
+sshKey: |
+  {{sshKey}}
+{% endif %}
```
durera, line 37: Linting ... You've lost the empty newline at the bottom here as part of this update.

ibm/mas_devops/roles/ocp_provision/README.md
```diff
@@ -241,6 +241,14 @@ The working directory that is used to perform the installation, it will contain
 - Environment Variable: `IPI_DIR`
 - Default Value: `~/openshift-install`
 
+### sshKey
```
durera, line 244: You're adding new capability, this needs to be raised as a [minor] update rather than a [patch] .. and as a result update the changelog in docs/changes.md and ibm/mas_devops/README.md

durera (changes requested):

ibm/mas_devops/roles/ocp_provision/README.md
vskhabani: @durera 12.2 would be the next release right ?

vskhabani:
@durera , I have updated the changelog in docs/changes.md and ibm/mas_devops/README.md

whitfiea (changes requested):
the version would be 12.3 not 12.2

whitfiea (approved):

whitfiea (approved):

PR #544, opened 2022-12-05
[patch] Fix storage class determination for MVI

alequint:
Before the fix, storage class to be used when installing MVI was being checked for every application running suite-app-install role. This PR modifies that in a way that, similar to what is implemented for `iot`, it only determine mvi class when installing mvi.


pseemung (commented):
Code switch from storage classes to actual application - visualinspection.

PR #562, opened 2022-12-13
[minor] Support FIPS option on FYRE infrastructure

HariPalleti:
## Description
MAS Devops Ansible Collection to support enabling FIPS when provisioning an OCP cluster
fyre environment provision will be configured by setting an environment variable `OCP_FIPS = yes/no`

## Related Issues
- https://github.ibm.com/wiotp/tracker/issues/11038

## How Has This Been Tested?
Have created quickburn cluster on fyre by providing environment variable `OCP_FIPS` and also ran without providing the env

### DIFF
--- a/ibm/mas_devops/roles/ocp_provision/defaults/main.yml
+++ b/ibm/mas_devops/roles/ocp_provision/defaults/main.yml
@@ -90,3 +90,5 @@ ipi_config_dir: "{{ ipi_dir }}/config/{{ cluster_name }}"
 
 ocp_installer_dir: "{{ ipi_dir }}/installer/{{ ocp_version }}"
 ocp_installer_exe: "{{ ipi_dir }}/installer/{{ ocp_version }}/openshift-install"
+
+ocp_fips: "{{ lookup('env', 'OCP_FIPS') | default('no', true) }}"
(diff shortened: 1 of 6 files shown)


ibm/mas_devops/roles/ocp_provision/defaults/main.yml
```diff
@@ -90,3 +90,5 @@ ipi_config_dir: "{{ ipi_dir }}/config/{{ cluster_name }}"
 
 ocp_installer_dir: "{{ ipi_dir }}/installer/{{ ocp_version }}"
 ocp_installer_exe: "{{ ipi_dir }}/installer/{{ ocp_version }}/openshift-install"
+
+ocp_fips: "{{ lookup('env', 'OCP_FIPS') | default('no', true) }}"
```
durera, line 94: Can we make this a boolean value, e.g.

```
ocp_fips: "{{ lookup('env', 'OCP_FIPS') | default('false', true) | bool }}"
```

Also need to update the README to add this new parameter to the documentation

ibm/mas_devops/roles/ocp_provision/templates/fyre/quick_burn.json.j2
```diff
@@ -17,5 +17,6 @@
             "http-keep-alive": "10s",
             "check": "10s"
         }
-    }
+    },
+    "fips":"{{ ocp_fips }}"
```
durera, line 21: - Use the new bool to convert from true/false to "yes"/"no" string specific for fyre.
- Should we not also add the same to `product_group.json.j2` as well?

durera (changes requested):
Also, this is a minor change rather than a patch -- we are adding new feature to the collection, please update the changelog accordingly.

durera (approved):

PR #593, opened 2023-01-09
[patch] Predict CRD updates

karol-czarnecki:
https://github.ibm.com/maximo/Asset-Health-Insight/issues/13018


lokesh-sreedhara (dismissed):

lokesh-sreedhara (dismissed):

lokesh-sreedhara:
@durera can u help review this PR and merge it into master?

ibm/mas_devops/roles/suite_app_config/vars/predict.yml
```diff
@@ -8,9 +8,10 @@ mas_app_cfg_retries: 30
 
 mas_app_settings_wml_instance_id: "{{ lookup('env', 'WML_INSTANCE_ID') | default('openshift', true) }}"
 mas_app_settings_wml_url: "{{ lookup('env', 'WML_URL') | default('https://internal-nginx-svc.ibm-cpd.svc:12443', true) }}"
-mas_app_settings_wml_version: "{{ lookup('env', 'WML_VERSION') }}"
+mas_app_settings_wml_version: "{{ lookup('env', 'WML_VERSION') | default('4.5', true) }}"
```
durera, line 11: Can we not set default here ... keep it a required input, unless there is a default that works for all versions of Predict ... put defaults that are suitable for the latest version in the playbooks instead (so they are easy to find).

durera (changes requested):

ibm/mas_devops/roles/suite_app_config/vars/predict.yml
lokesh-sreedhara: done. changes made as requested.

ibm/mas_devops/roles/suite_app_config/vars/predict.yml
```diff
@@ -11,6 +11,7 @@ mas_app_settings_wml_url: "{{ lookup('env', 'WML_URL') | default('https://intern
 mas_app_settings_wml_version: "{{ lookup('env', 'WML_VERSION') }}"
 mas_app_settings_predict_deployment_size: "{{ lookup('env', 'PREDICT_DEPLOYMENT_SIZE') | default('small', true) }}"
 
+mas_app_settings_watsonstudio_project_id: "{{ lookup('env', 'WS_PROJECT_ID') | default('', true) }}"
```
durera, line 14: Env var lookups should all be in defaults, not vars .. but seeing as there are existing problems in this role in this area will not reject this ... but next update to this role we need to sort out what should be in vars and what should be in defaults.

(The reason is that env var lookups in vars files will override any parameters passed in on command line, or via a playbook .. making env vars the **only** way to set the variable)

PR #595, opened 2023-01-10
[patch] Add skip_compatibility_check for upgrade error path test

liyongc:
For suite_upgrade and suite_app_upgrade, it will auto do compatibility check.
But for upgrade test, we need to do the error path validation and check if the non-supported upgrade path is checked by mas core and app operator. 
So here I add Env var "skip_compatibility_check" if we want to skip compatibility check for upgrade-test error path test.
By default, skip_compatibility_check will be set "False" and it means it will perform compatibility_check.

- [x] Test against MAS CLI upgrade-test pipeline (https://github.com/ibm-mas/cli/pull/130) - in progress (@alequint )

**Update** (Alex): @liyongc  I have moved skip condition to skip compatibility check only instead of remove all the check compatibility tasks for suite and app upgrades. Reason for that is because among those tasks we have also, besides compatibility, some checkings to see if are already on the target channel, etc, that we can keep. Pls review it.

- [x] Update Jan 19th: suite-app-upgrade is failing when no target channel is defined, even if skip is true. Need to review criteria. Work In progress and PR is back to DRAFT

**Update** (Alex): Jan 23th: PR is back to the game... fixed the issue. When app channel is not informed (informing, in upgrade pipeline, that we don't want to upgrade that app, role now works as follow:

- When `SKIP_COMPATIBILITY_CHECK`=false

image

- When `SKIP_COMPATIBILITY_CHECK`=true

image


alequint:
@liyongc let me include this PR in my version of CLI... I think I need that to be able to upgrade my channel to a development channel, as it is the plan for upg-stable tests we are doing. I will move to draft till my upgrade test is done, so we can put some evidences of test here when this capability is used in the pipeline

alequint:
Latest update: https://ibm-watson-iot.slack.com/archives/C04A3707CLS/p1673697434342509?thread_ts=1673607634.489709&cid=C04A3707CLS

durera (changes requested):
Just needs a readme update to document the new skip_compat_check variable (in both updated roles)

liyongc:
@durera I just updated the README with the parameter description, thanks. @alequint, fyi

durera (approved):

PR #599, opened 2023-01-11
[patch] Make ocp_upgrade flexible enough to upgrade any ocp_version

andrercm:
I'm not sure why but when i implemented this role a long time ago, I haven't though about making it flexible enough so it can be used to upgrade any ocp version... fixing this to be better fit for our upcoming upgrade testing...


andrercm:
This is ready for review and merge

durera (approved):

durera (approved):

PR #610, opened 2023-01-17
[patch] adding flag to enhance dns security

JonahLuckett:
- Added `cis_enhanced_security` flag (default: false) so that users can choose to use the new or old cis suitedns methods
- Removed need to log into ibmcloud in the terminal to run the enhanced version
- Updated readMe to include details on what the enhanced security option means

See [relevant slack thread](https://ibm-watson-iot.slack.com/archives/C0195MVCEUD/p1673989120058799?thread_ts=1673986757.612829&cid=C0195MVCEUD)

### DIFF
--- a/ibm/mas_devops/roles/suite_dns/README.md
+++ b/ibm/mas_devops/roles/suite_dns/README.md
@@ -121,6 +121,26 @@ Role Variables - IBM Cloud Internet Services DNS Integration
 - Environment Variable: `CIS_SUBDOMAIN`
 - Default: None
 
+### cis_enhanced_security
+Set this to true to enable the enhanced CIS DNS integration security. See below for details.
+
+- Optional
+- Environment Variable: `CIS_ENHANCED_SECURITY`
+- Default: false
+
+
+Role Variables - Enhanced IBM CIS DNS Integration Secruity
+------------------------------------------------------------
+- Enable WAF firewall disabling rules that affect MAS application functionalities
+- Enable Proxy for DNS entries
+- Ensure there are no wildcard DNS entry in CIS
+- Create Edge Certificates in CIS instance
+
+### ibmcloud_api_key
+
+- **Required** if `cis_enhanced_security` is set to `true`
+- Environment Variable: `IBMCLOUD_API_KEY`
+
 ### cis_waf
 
 - Optional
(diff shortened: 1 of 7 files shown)


ibm/mas_devops/roles/suite_dns/README.md
```diff
@@ -121,6 +121,26 @@ Role Variables - IBM Cloud Internet Services DNS Integration
...
+
+
+Role Variables - Enhanced IBM CIS DNS Integration Secruity
+------------------------------------------------------------
+- Enable WAF firewall disabling rules that affect MAS application functionalities
+- Enable Proxy for DNS entries
+- Ensure there are no wildcard DNS entry in CIS
+- Create Edge Certificates in CIS instance
+
+### ibmcloud_api_key
```
durera, line 139: We already have `ibmcloud_apikey` .. which is used when provisioning clusters ... what is the purpose of this variable?

If this is passed into an automated process that communicates with CIS can we use the existing `cis_apikey`?  If this is used as one-time (when the role runs) authentication to IBM Cloud, can we use ibmcloud_apikey, so that we don't have ibmcloud_apikey and ibmcloud_api_key.

ibm/mas_devops/roles/suite_dns/defaults/main.yaml
```diff
@@ -41,6 +41,11 @@ cis_email: "{{ lookup('env', 'CIS_EMAIL') }}"
 cis_apikey: "{{ lookup('env', 'CIS_APIKEY') }}"
 cis_crn: "{{ lookup('env', 'CIS_CRN') }}"
 cis_subdomain: "{{ lookup('env', 'CIS_SUBDOMAIN') }}"
+cis_enhanced_security: "{{ lookup('env', 'CIS_ENHANCED_SECURITY') | default(false, false) }}"
+
+# Enhanced IBM CIS DNS Integration Secruity
+# -----------------------------------------------------------------------------
+ibmcloud_api_key: "{{ lookup('env', 'IBMCLOUD_API_KEY') | default(true, true) }}"
```
durera, line 48: See previous comment about existing ibmcloud_apikey var/env var

ibm/mas_devops/roles/suite_dns/README.md
```diff
@@ -121,6 +121,26 @@ Role Variables - IBM Cloud Internet Services DNS Integration
 - Environment Variable: `CIS_SUBDOMAIN`
 - Default: None
 
+### cis_enhanced_security
+Set this to true to enable the enhanced CIS DNS integration security. See below for details.
```
durera, line 125: Can you add something that explains what this actually gives the user .. e.g. a link to the CIS docs related to enhanced security, someone reading our docs should be presented a simple avenue to find out more about this option.

durera (changes requested):

durera (approved):

PR #629, opened 2023-01-27
[patch] Replace use of community.kubernetes with kubernetes.core

racree:
community.kubernetes has been deprecated and we need to use kubernetes.core instead.


racree:
verified fvt tests pass

durera (approved):

PR #649, opened 2023-02-06
updates ingress check

tcskill:
Updates the ingress cert check to cover IPI installs of OCP such as in new TechZone clusters


tcskill:
@durera fyi - this is what we discussed in the channel on supporting the new techzone clusters. Tested and passed IPI / OCP install fine.

durera (approved):

PR #671, opened 2023-02-17
[patch] use Discovery Advanced setting for assist activation

liyongc:
We change the assist activation page to use advanced setting for discovery related configuration, so we need to change the below part to let assist activation to use the advanced setting.
1. Update role `cp4d_service` to save the discovery instance url to mas config folder during discovery activation.
2. The info will be used by  role `suite_app_config` to combined the new Assist workspace CR spec.


durera (changes requested):

whitfiea:
@liyongc can you have a look at the review comments from David?

PR #687, opened 2023-03-02
 [patch] Internal db2 support for manage 

padmankosalaram:
Previous PR detail
Fix DB21015E error during db2 prepare script execution - https://github.com/ibm-mas/ansible-devops/pull/676


durera (changes requested):
We are not going to introduce ignore errors to the code ... things need to work reliably, ignoring errors is not an option; fix the underlying issue causing the error.

padmankosalaram:
Closing this as the fix is available in https://github.com/ibm-mas/ansible-devops/pull/688

PR #696, opened 2023-03-13
[patch] Fix attachments config for Health Standalone

andrercm:
make the attachments and bim configurations to be flexible for it can be reused for health standalone


andrercm:
above is already tested in a health standalone environment and ready to be merged

durera (approved):

PR #703, opened 2023-03-16
[patch] Fix mqttBroker storage class and size configuration for IoT

IanBoden:
Without a mqttBroker storage class and size IOT t-shirt size is not passed to the mbgx CR so by setting a default will ensure we don't get stuck with the IOT wanting small/medium and the mbgx using dev

Also changed the contributing to match the makefile targets


ibm/mas_devops/roles/suite_app_install/vars/defaultspecs/iot.yml
```diff
@@ -9,6 +9,10 @@ mas_app_spec:
   settings:
     deployment:
       size: "{{ mas_app_settings_iot_deployment_size }}"
+    messagesight:
+      storage:
+        class: "{{ mas_app_settings_iot_mqttbroker_pvc_storage_class }}"
+        size: "{{ mas_app_settings_iot_mqttbroker_pvc_size }}"
```
durera, line 15: As implemented wouldnt this lead to a scenario where if I set:

mas_app_settings_iot_deployment_size: dev

I would end up with:

```
  settings:
    deployment:
      size: "{{ mas_app_settings_iot_deployment_size }}"
    messagesight:
      storage:
        class: ""
        size: 100Gi
```

Think we need a `{% if size != dev %}` style block around this new section so we don't specify storage for a dev install - which doesn't use PVCs IIRC?  Something like that anyway - similar to the way the auto storage class selection doesn't do anything if size == dev?

durera (changes requested):

PR #712, opened 2023-03-21
[patch] Fix AWS cluster storage support

pseemung:
1. Changes to ensure that storage class provided by the user gets picked up and used. This was causing a problem because the variables were overwritten by another setup which made the PVC to use the default settings storage class. Since aws did not have an ibmc-file-gold, it would always break the pipeline.
2. Update driver version default values using the latest MVI supported available in nvidia repository. 
3. Update the default gpu channel because there was a new release and the stable channel points to that but MVI is not supporting that yet. 
4. Documentation for the above changes.


ibm/mas_devops/roles/suite_app_install/README.md
```diff
@@ -107,11 +107,10 @@ Optional. Defines what plan will be used in application install.
 Role Variables - Visual Inspection Configuration
 -------------------------------------------------------------------------------
 ### mas_app_settings_visualinspection_storage_class
-Storage class used for user data. This must support ReadWriteOnce
+Optional - Storage class used for user data. This must support ReadWriteOnce
```
maulik-modi22, line 110: must be ReadWriteMany

ibm/mas_devops/roles/suite_app_install/README.md
```diff
@@ -107,11 +107,10 @@ Optional. Defines what plan will be used in application install.
...
 
-- Optional
 - Environment Variable: `MAS_APP_SETTINGS_VISUALINSPECTION_STORAGE_CLASS`
-- Default: Defaults to `ibmc-file-gold` if the storage class is available in the cluster.
+- Default: Auto-selected from storage classes installed in the cluster.
```
maulik-modi22, line 113: MVI EFS requires uid:`999` due to mongoldb requirement and Cluster EFS requires uid:`0';
Auto selection would 100% fail MVI would fail with `Could not bind PVC`

```kind: StorageClass
apiVersion: storage.k8s.io/v1
metadata:
  name: mvi-efs
  uid: 85441182-8403-4793-a34c-04ae0b0e1350
 provisioner: efs.csi.aws.com
parameters:
  basePath: /mvi
  directoryPerms: '777'
  fileSystemId: fs-00c7e21665962211a
  gid: '0'
  provisioningMode: efs-ap
  uid: '999'
reclaimPolicy: Delete
volumeBindingMode: Immediate```

Cluster EFS requires aid:'0'
```kind: StorageClass
apiVersion: storage.k8s.io/v1
metadata:
  name: efs
  uid: 85441182-8403-4793-a34c-04ae0b0e1350
 provisioner: efs.csi.aws.com
parameters:
  basePath: /rosa
  directoryPerms: '777'
  fileSystemId: fs-00c7e21665962211a
  gid: '0'
  provisioningMode: efs-ap
  uid: '0'
reclaimPolicy: Delete
volumeBindingMode: Immediate```

File: ibm/mas_devops/playbooks/oneclick_add_visualinspection.yml [hunk 3]
```diff
@@ -12,7 +12,6 @@
     # mas_app_id can be set to "health" to install manage in the "Health standalone" mode from this same playbook.
```
maulik-modi22, line 12: not relevant comment

ibm/mas_devops/roles/suite_app_install/tasks/visualinspection.yml
```diff
@@ -22,13 +22,12 @@
...
-  set_fact:
-    mas_app_settings_visualinspection_storage_class: "{{ lookup_storageclasses | ibm.mas_devops.defaultStorageClass(default_storage_classes_rwx_nogid) }}"
 
-- name: Assert that storage class has been defined
-  assert:
-    that: mas_app_settings_visualinspection_storage_class is defined and mas_app_settings_visualinspection_storage_class != ""
-    fail_msg: "mas_app_settings_visualinspection_storage_class must be defined"
+  vars:
+    # ROKS, OCS, AWS, Azure
+    supported_storage_classes: [ibmc-block-gold, ocs-storagecluster-ceph-rbd, gp2, efs, managed-premium]
```
andrercm, line 28: any reason to not use `ibm/mas_devops/common_vars/default_storage_classes.yml` to maintain the defaulted storage classes for MVI? It's probably easier to maintain this code in the future if we keep all the defaulted storage class in a centric place rather than keep it isolated in each playbook/role.

You could probably just create a new property specific to MVI there if that's the case.

docs/changes.md
```diff
@@ -1,6 +1,7 @@
 ## Changes
 Note that links to pull requests prior to public release of the code (4.0) direct to IBM GitHub Enterprise, and will only be accessible to IBM employees.
 
+- `12.15` Add additional AWS storage support for MVI playbook ([#712](https://github.com/ibm-mas/ansible-devops/pull/712))
```
andrercm, line 4: Can you also update change logs in `ibm/mas_devops/README.md` ?

andrercm (changes requested):

docs/changes.md
pseemung: updated change log.

ibm/mas_devops/roles/suite_app_install/README.md
pseemung: ReadWritemany has been updated.

Not adding MVI EFS requirement as it is not required for MAS 8.10.  Customer can select EFS storage class using MVI 8.8.

ibm/mas_devops/roles/suite_app_install/tasks/visualinspection.yml
pseemung: Worked with Caroline to use the default stoage class for MVI.

pseemung (commented):
Please review changes.

ibm/mas_devops/README.md
```diff
@@ -5,7 +5,7 @@
 
 ## Change Log
 Note that links to pull requests prior to public release of the code (4.0) direct to IBM GitHub Enterprise, and will only be accessible to IBM employees.
-
+- `13.2` Add additional AWS storage support for MVI playbook ([#712](https://github.com/ibm-mas/ansible-devops/pull/712))
```
durera, line 8: The new line is needed before the list starts, or this won't render correctly .. also the other changelog in docs/changes.md also needs the same update

ibm/mas_devops/roles/suite_app_install/tasks/main.yml
```diff
@@ -15,17 +15,16 @@
 - name: Load variables
   include_vars: "vars/{{ mas_app_id }}.yml"
 
-# 2.1 Load default application spec, if none is provided by user
-- name: Load default application spec
-  when: mas_app_spec is not defined or mas_app_spec == ""
-  include_vars: "vars/defaultspecs/{{ mas_app_id }}.yml"
-
-# 2.2  Run Application Specific Tasks
+# 2.1  Run Application Specific Tasks
```
durera, line 18: What's the reason for re-ordering these?  Have you verified that neither of the application specific tasks require the values laoded in the vars file, becuase these vars will no longer be available when tasks/$appId.yml runs now?

durera (changes requested):

ibm/mas_devops/roles/suite_app_install/tasks/main.yml
caroazad: It was reordered because there the storage variable was overwritten by the second one at the time so the mas_app_spec became {} and resulted to using the default specs. @pseemung Maybe you could explain better.

pseemung: @andrercm assisted me because I could not get the storage classes to work properly.  The problem with the original order was that I defined the custom storage class for mvi after the` mas_app_spec` was already set.  The order was changed because it defines the default storage classes first, then you call the step that sets the mas_app_spec if that’s not defined.

If you believe this is not correct, please let me know how I would resolve this scenario.

durera: That's fine :+1: ... but my question remains unanswered:

> Have you verified that neither of the application specific tasks require the values laoded in the vars file

The focus of this PR is MVI, but IoT is also impacted by this change, has it been tested that nothing in the IoT app-specific task required those vars to have been set up before it ran?

caroazad: Oh I see. You want to make sure that IoT is working fine even with this change. @pseemung Do you have time today to test with IoT?

durera (approved):

PR #718, opened 2023-03-22
[patch] add work-around to pass predict version check

liyongc:
suite_upgrade fail to pass due to predict operator **status.version** is used for now, instead it should **status.versions** as other application, So here temporarily to bypass the issue and get the upgrade flow works for all apps.


lokesh-sreedhara:
This PR on Predict operator fixes this issue 
https://github.ibm.com/maximoappsuite/ibm-mas-predict/pull/320 

we might not need this PR anymore once these changes are in

PR #722, opened 2023-03-23
[patch] Fix for suite_dns security enhancements

andrercm:
This fixes the `cis_dns_entries.py` module to include parameters needed by SRE to allow CIS to enable proxy and few other security related items.

1. copied the `cis_dns_entries.py` module from SRE collection into our public collection. SRE originally used our own module as base for theirs, now we're bringing back their additional changes into ours :)
2. removed the need for an extra env var `IBMCLOUD_APIKEY` that we added last time we changed this code as part of the SRE merge epic, as we already have `CIS_APIKEY` and we don't need two keys for the same thing (make calls to CIS), the original `CIS_APIKEY` one is good enough.


andrercm:
@CAIOFCP @durera @JonahLuckett can you help me review and merge this? I really need this merged to enable security enhancements to be added as part of our fvt/saas pipeline, and also to more efficiently provision environments using the pipelines with same security settings as SRE uses.

andrercm:
I tested these changes and now I can have the proxy enabled in CIS for my cname entries:
image

JonahLuckett:
Happy to approve but I'm not a code owner so can't merge yet

JonahLuckett (dismissed):

PR #730, opened 2023-03-25
[patch] Multiple fixes for CIS security settings

andrercm:
This PR fixes more problems found during the E2E tests while enabling security settings for CIS:

- Standardizes the way we lookup for DNS entries for both basic and security enabled paths.
- Fixes the `cis_dns_entries.py` to use `cis_proxy` flag - it was being set but never used in the python module (hardcoded to `true`)!
- Aligns the report bundle server name and sub route in `Manage` workspace activation to be `rpt` so it matches with the DNS entry created by `suite_dns` role and the corresponding Manage route created for report workload.
- Fixes the setup of the DNS entries to use `mas_workspace_id` as parameter for the entries that requires workspace in it - it was hardcoded to `main`!
- Enables the Edge certificates to be deleted or overwritten based on the `mas_instance_id`, this way we allow multiple Edge Certificates to be deleted or created in a shared CIS instance. The existing code checks if there is any Edge cert in the CIS instance for the domain, and if there is, it picks the first item (it doesn't care which one it is - i think it always expect to only exist one edge cert per CIS instance...) and sets this to be the certificate Id to be deleted. With this fix, it will get the certificate id based on the edge certificates generated to an specific `mas_instance_id`, so we know for sure that this is the right one to be deleted and recreated, and not mess with any other edge cert that might exist on the CIS instance.


andrercm:
@CAIOFCP @JonahLuckett i'll need help specially for you on reviewing it as you guys are also familiar with the security settings for it... I have tested multiple scenarios where I created dns entries with our without security settings enabled, enabled/disabled proxy, and deleted/recreated edge certificates, hopefully that's the remaining piece of work we'll have to do to have this role flexible enough for our teams (dev/test/sre) to use and also customers.

ibm/mas_devops/roles/suite_app_config/vars/manage.yml
```diff
@@ -107,9 +107,9 @@ mas_app_settings_server_bundles:
         isDefault: false
         isMobileTarget: false
         isUserSyncTarget: false
-        name: report
+        name: rpt
```
durera, line 110: Would this be a breaking change for anyone running the install over the top of an existing install?

andrercm:
@durera The changes above should go all together with cli changes from this PR: https://github.com/ibm-mas/cli/pull/213

andrercm:
Just reviewed that with @CAIOFCP , all looks good in his view.

PR #731, opened 2023-03-27
[patch] Fix Manage one-click playbook documentation

sbudhira:


durera (approved):
Please write complete PRs in future, normally I would reject this as it has no description, but I'm on vacation so letting it slide to get it into today's release.  It doesn't need to be war and peace, just needs a single sentence explaining what's been changed.

PR #732, opened 2023-03-27
[patch]fixed mongodb role for action 'none'

Ashish-Mahawal1:
Issue: 1. Playbook uninstall-core was failing when given value `none` for skipping any dependency uninstall

https://ibm-watson-iot.slack.com/archives/C040TETRV2T/p1679677295384159

Changes :
1. Changed default for mongodb_action to `deprovision` in playbook `uninstall-core`
2. Added a condition to skip any mongodb action when `mongodb_action` is set to `none`

### DIFF
--- a/ibm/mas_devops/playbooks/uninstall_core.yml
+++ b/ibm/mas_devops/playbooks/uninstall_core.yml
@@ -6,7 +6,7 @@
     # Flip the default actions for each role to "uninstall"
     cluster_monitoring_action: "{{ lookup('env', 'CLUSTER_MONITORING_ACTION') | default('uninstall', true) }}"
     sls_action: "{{ lookup('env', 'SLS_ACTION') | default('uninstall', true) }}"
-    mongodb_action: "{{ lookup('env', 'MONGODB_ACTION') | default('uninstall', true) }}"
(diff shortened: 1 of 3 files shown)


durera (changes requested):
MongoDb actions should be `install` / `uninstall` to be consistent with all other `xxx_action` options.

Can we fix this here to align on the terminology .. I don't really mind what the terminology is .. add/remove, provision/deprovision, or install/uninstall - but we have used install/uninstall the most so let's stick to that and update mongo role to bring it into line.

Ashish-Mahawal1:
> MongoDb actions should be `install` / `uninstall` to be consistent with all other `xxx_action` options.
> 
> Can we fix this here to align on the terminology .. I don't really mind what the terminology is .. add/remove, provision/deprovision, or install/uninstall - but we have used install/uninstall the most so let's stick to that and update mongo role to bring it into line.

Done all requested changes ,please re-review thanks

andrercm (approved):
LGTM

andrercm:
@durera since you helped review it at first... do you have any further considerations?

whitfiea (approved):

PR #748, opened 2023-04-04
[patch] Fixes for MongoCfg generation with existing IBMCloud MongoDb

andrercm:
This PR fixes a couple of things in mongodb role:

- Adjusts the prefixes for defaulted vars to start with `mongodb_`
- Fixes the MongoCfg config to create the Mongo Credentials secret along with the MongoCfg for IBM Cloud Mongo provisioning.
- Allow passing `IBM_MONGO_ADMIN_PASSWORD` so could generate a MongoCfg that works for an existing IBM Cloud Mongo instance if we don't want to provision a new Mongo.
- Creates a new var `IBM_MONGO_RESOURCEGROUP` which allow us to be more flexible to where we want the IBM Cloud Mongo to be provisioned or to be found, specially if we want it to exist in a different resource group as the cluster is located (which uses IBMCLOUD_RESOURCEGROUP).
- Updates defaulted IBM Cloud Mongo version to 4.4 instead of 4.2 since the latter is quite old.


andrercm:
Tested with fvt-personal and with the above PR, we can now setup both SLS and MAS to use a new or existing IBM Cloud Mongo database along with the pipeline:

image

Just need to export the following to the pipeline:

```
    # IBM Cloud MongoDB
    # -------------------------------------------------------------------------
    - MONGODB_PROVIDER=ibm
    - IBM_MONGO_NAME=mongodb-test
    - IBM_MONGO_RESOURCEGROUP=MASIVT
    - IBM_MONGO_REGION=us-east
    - IBM_MONGO_ADMIN_PASSWORD=< your admin password for mongo instance > 

    # SLS - point to existing IBM MongoDB
    # -------------------------------------------------------------------------
    - SLS_NAMESPACE=ibm-sls-$MAS_INSTANCE_ID  
    - SLS_MONGODB_CFG_FILE=/workspace/configs/ibmcloud-mongo-$IBM_MONGO_NAME-cfg.yml
```

durera (approved):

PR #752, opened 2023-04-05
[major] Switch to ARTIFACTORY_TOKEN

caroazad:
Update all files to change artifactory_apikey to artifactory_token. We need this because airgap dev is blocked by it due to the name change and it's failing. Thank you

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,8 +7,8 @@
 build:
 	ansible-galaxy collection build --output-path . ibm/mas_devops --force
 install:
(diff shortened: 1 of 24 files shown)


andrercm (changes requested):
This seems like a disruptive change at least for development teams, so I think as enough motivation to become a minor release, so we need to add a change logs for it to declare and make it clear when this was done and released.

durera (dismissed):

durera (approved):

PR #756, opened 2023-04-06
[patch] Update all defaults for MAS 8.10

durera:
- Update all default channels in install playbooks for MAS 8.10 release
- Update default catalog choice in mirror playbooks to `v8-230328-amd64`


aroute:
This change is not showing up in cli/docker version of 3.20.0. The core playbook still shows MAS 8.9.x.

PR #765, opened 2023-04-12
[patch] support appconnect and other dependencies

caroazad:
- [x]  setup mirroring for appconnect
- [x] - allow excluded images in common services mirroring that are depended on by appconnect
- [x] - Install appconnect fully including dashboard in airgap environment as a test


ibm/mas_devops/common_vars/casebundles/v8-230414-amd64.yml
```diff
@@ -25,6 +26,9 @@ mas_core_version:
 mas_assist_version:
   8.9.x: 8.6.3    # Updated
   8.10.x: 8.7.0   # No Update
+mas_hputilities_version:
+  8.9.x: 8.5.1    # Updated
+  8.10.x: 8.6.0   # Updated
```
whitfiea, line 31: Is this PR also going to include hpu to be installed using airgap? If so can we update the PR description and PR title so it says that as well. It should also be a minor rather than patch and have changelog updates

ibm/mas_devops/playbooks/mirror_add_hputilities.yml
```diff
@@ -17,7 +17,7 @@
     mas_channel: "{{ lookup('env', 'MAS_CHANNEL') | default ('8.10.x', True) }}"
     mirror_mode: "{{ lookup('env', 'MIRROR_MODE') | default ('direct', True) }}"
 
-    # mirror_appconnect: "{{ lookup('env', 'MIRROR_APPCONNECT') | default ('True', True) | bool }}"
+    mirror_appconnect: "{{ lookup('env', 'MIRROR_APPCONNECT') | default ('True', True) | bool }}"
     # mirror_cp4d: "{{ lookup('env', 'MIRROR_CP4D') | default ('True', True) | bool }}"
     # mirror_wsl: "{{ lookup('env', 'MIRROR_WSL') | default ('True', True) | bool }}"
```
whitfiea, line 22: If hpu also needed cp4d and wsl and these are not airgap installable yet, then I guess we can't claim that hpu is installable via airgap?

whitfiea (changes requested):

ibm/mas_devops/playbooks/mirror_add_hputilities.yml
caroazad: Yes right!

ibm/mas_devops/common_vars/casebundles/v8-230414-amd64.yml
whitfiea: You would always need to mirror core (even if you did it long ago you can still run it again) so I don't think we need to add it into mirror-add-hputilities as well. If mirror-core was optional then I would say yes, but it isn't

Issue #723, opened 2023-03-24
ocs role does not complete its install on OCP 4.11 or 4.12


jeancrg:
Tried ocs role in OCP 4.11 and 4.12 on Fyre (Quick Burn) and the install does not complete. It works fine on 4.10 and previous but on 4.11/4.12 it doesn`t. It remains stuck in the ibm.mas_devops.ocs until it fails after all the retries.

```
21:46:29 TASK [ibm.mas_devops.ocs : ocs/storage : Create OSC Cluster] *******************
21:46:29 Friday 24 March 2023  00:46:29 +0000 (0:00:21.404)       0:12:11.397 ********** 
[...]

Issue #522, opened 2022-11-18
the device path of the ocs role should be customizable


hughjfchen:
Currently, the `DevicePath` of the `ocs` role is hard coded as `/dev/vdb`, however in some situation this is not the case
(for example, it may be `/dev/sdb` in a mearmetal machine) so it should be customizable. Suggest add a variable in the 
template for this.

andrercm:
This seems like a quick winner for our list of backlogs, i'll include it in the backlog for the next upcoming work

PR #766, opened 2023-04-13
[minor] Update ocs role to use odf operator for OCP 4.11+ and provide devicePaths option

whitfiea:
This PR updates the local storage operator (lso) and the openshift storage operator (odf) so it can be used in OCP 4.11 and onwards. In OCP 4.11 the operator changed name from ocs (Openshift Container Storage) to odf (OpenShift Data Foundation). The resources it controls (StorageCluster) remained the same so this was just an update to use the newly named operator. A difference for the local storage operator (lso) is that the channel subscription for the operator changed from being the OCP version to be "stable".

The logic is now:
If OCP_RELEASE is 4.11 or greater then use lso channel of `stable` and the `odf` operator rather than `ocs`
If OCP_RELEASE is less than 4.11 then use lso channel of `the OCP_RELEASE` and the `ocs` operator 

I have also added the ability to provide the device path to be used in the local volume that the local storage operator uses. The local volume still has to be block, but the device path can now be set if different to `/dev/vdb` based on feedback in issue https://github.com/ibm-mas/ansible-devops/issues/522

I have tested all this in Fyre using OCP versions 4.9, 4.10, 4.11 and 4.12 and it works as expected.

This resolves https://github.com/ibm-mas/ansible-devops/issues/723 and resolves https://github.com/ibm-mas/ansible-devops/issues/522


ibm/mas_devops/roles/ocs/defaults/main.yml
```diff
@@ -1,3 +1,9 @@
 ---
 # We need to know what the OCP_RELEASE is to know what channel to target the subscriptions at
-ocp_release: "{{ lookup('env', 'OCP_RELEASE') | default('4.10', True) }}"
+ocp_release: "{{ lookup('env', 'OCP_RELEASE') | default('4.12', True) }}"
```
durera, line 3: While you're making changes here could you fix (remove) this OCP_RELEASE stuff ...

```
oc get clusterversion version
NAME      VERSION   AVAILABLE   PROGRESSING   SINCE   STATUS
version   4.12.7    True        False         36h     Cluster version is 4.12.7
```

User should not need to tell us the cluster release ... we can easily determin it from the clusterversion resource

andrercm (dismissed):

durera (dismissed):

andrercm (approved):
LGTM

PR #769, opened 2023-04-14
[patch] Fix Manage JMS related secret names to be consistent with the pattern that MAS UI generates

andrercm:
Currently when Manage is activated with `mas_app_settings_server_bundles_size : jms`, by default we create custom secret names that contain the xml config to setup JMS pre-reqs for queues and linked them in the ManageWorkspace CR.

These secrets are named:
```
main-manage-additional-server-config-no-activation
main-manage-additional-server-config
main-manage-jms-additional-server-config
```

The problem here is that when someone make any change to Manage Workspace via UI, the UI checks that these secret are setup with a different secret name pattern than it expects and renames all these secrets to following pattern:

```
main-manage-d--sb0--asc--sn
main-manage-d--sb1--asc--sn
main-manage-d--sb2--asc--sn
main-manage-d--sb3--asc--sn
main-manage-d--sb4--asc--sn
```

This causes a mismatch in the JMS configuration as the ManageWorkspace CR does not look for the right secrets anymore once this happens. 
This PR aligns the secret names created by the Manage JMS setup automation to what the MAS UI expects, to avoid messing up with the Manage bundles server configurations.

### DIFF
--- a/ibm/mas_devops/roles/suite_app_config/vars/health.yml
+++ b/ibm/mas_devops/roles/suite_app_config/vars/health.yml
@@ -13,8 +13,6 @@ mas_app_settings_demodata: "{{ lookup('env', 'MAS_APP_SETTINGS_DEMODATA') | defa
 mas_app_settings_tablespace: "{{ lookup('env', 'MAS_APP_SETTINGS_TABLESPACE') | default('MAXDATA', true)}}"
 mas_app_settings_indexspace: "{{ lookup('env', 'MAS_APP_SETTINGS_INDEXSPACE') | default('MAXINDEX', true)}}"
 
-server_bundles_jms_add_server_config_secret_name: "{{ mas_workspace_id }}-manage-jms-additional-server-config"
-server_bundles_add_server_config_secret_name: "{{ mas_workspace_id }}-manage-additional-server-config"
 mas_app_settings_persistent_volumes_flag: "{{ lookup('env', 'MAS_APP_SETTINGS_PERSISTENT_VOLUMES_FLAG') | default('false', true) }}"
 mas_app_settings_server_bundles_size: "{{ lookup('env', 'MAS_APP_SETTINGS_SERVER_BUNDLES_SIZE') | default('dev', true)  }}"
 mas_app_settings_server_bundles:
@@ -60,7 +58,7 @@ mas_app_settings_server_bundles:
   jms:
     serverBundles:
       - additionalServerConfig:
-          secretName: "{{ server_bundles_add_server_config_secret_name }}"
+          secretName: "{{ mas_workspace_id }}-{{ mas_app_id }}-d--sb0--asc--sn"
         bundleType: ui
         isDefault: true
         isMobileTarget: true
@@ -69,7 +67,7 @@ mas_app_settings_server_bundles:
         replica: 1
         routeSubDomain: ui
       - additionalServerConfig:
-          secretName: "{{ server_bundles_add_server_config_secret_name }}"
+          secretName: "{{ mas_workspace_id }}-{{ mas_app_id }}-d--sb1--asc--sn"
         bundleType: mea
         isDefault: false
         isMobileTarget: false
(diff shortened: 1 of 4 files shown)


andrercm:
Not sure if worths adding a step to clean up potential old/existing jms secrets while replacing it with the new secret names... it would be simple as just adding an extra step after the new secrets are created:

```
  - name: Delete old Manage app JMS additional server config secrets that uses old naming pattern
    kubernetes.core.k8s:
      state: absent
      api_version: v1
      kind: Secret
      name: "{{ item }}"
      namespace: "mas-{{ mas_instance_id }}-manage"
      wait: yes
      wait_timeout: 120
    with_items:
      - "{{ mas_workspace_id }}-{{ mas_app_id }}-additional-server-config"
      - "{{ mas_workspace_id }}-{{ mas_app_id }}-additional-server-config-no-activation"
      - "{{ mas_workspace_id }}-{{ mas_app_id }}-jms-additional-server-config"
```

Any thoughts around that?

ibm/mas_devops/roles/suite_app_config/vars/health.yml
```diff
@@ -13,8 +13,6 @@ mas_app_settings_demodata: "{{ lookup('env', 'MAS_APP_SETTINGS_DEMODATA') | defa
 mas_app_settings_tablespace: "{{ lookup('env', 'MAS_APP_SETTINGS_TABLESPACE') | default('MAXDATA', true)}}"
 mas_app_settings_indexspace: "{{ lookup('env', 'MAS_APP_SETTINGS_INDEXSPACE') | default('MAXINDEX', true)}}"
 
-server_bundles_jms_add_server_config_secret_name: "{{ mas_workspace_id }}-manage-jms-additional-server-config"
-server_bundles_add_server_config_secret_name: "{{ mas_workspace_id }}-manage-additional-server-config"
```
whitfiea, line 17: Would it be easier to set these as the new values rather than remove them and have the new values in mulitple places? i.e. why not change ```server_bundles_add_server_config_secret_name: "{{ mas_workspace_id }}-manage-additional-server-config"``` to be ```server_bundles_add_server_config_secret_name: "{{ mas_workspace_id }}-{{ mas_app_id }}-d--sb0--asc--sn``` ?

whitfiea: It would also make it easier to know what those are as `-d--sb0--asc--sn` means nothing to me.

whitfiea (changes requested):

whitfiea:
> Not sure if worths adding a step to clean up potential old/existing jms secrets while replacing it with the new secret names... it would be simple as just adding an extra step after the new secrets are created:
> 
> ```
>   - name: Delete old Manage app JMS additional server config secrets that uses old naming pattern
>     kubernetes.core.k8s:
>       state: absent
>       api_version: v1
>       kind: Secret
>       name: "{{ item }}"
>       namespace: "mas-{{ mas_instance_id }}-manage"
>       wait: yes
>       wait_timeout: 120
>     with_items:
>       - "{{ mas_workspace_id }}-{{ mas_app_id }}-additional-server-config"
>       - "{{ mas_workspace_id }}-{{ mas_app_id }}-additional-server-config-no-activation"
>       - "{{ mas_workspace_id }}-{{ mas_app_id }}-jms-additional-server-config"
> ```
> 
> Any thoughts around that?

Yeah I think we should do that as well. Could these old secrets contain anything that a user might set or would that always be changed via the UI (so would be the new names anyway)

whitfiea:
Actually, thinking about this more. If an existing customer had modified the JMS settings (and so has the new nameed secrets already). Will this change wipe over those modifications?

andrercm:
> Actually, thinking about this more. If an existing customer had modified the JMS settings (and so has the new nameed secrets already). Will this change wipe over those modifications?

If they rerun this role for whatever reason and `export MAS_APP_SETTINGS_SERVER_BUNDLES_SIZE='jms'` then yes, it would override the secret content if the secret names matches.

andrercm:
I'd think this would always be changed via UI has it's just easier for customers?

andrercm:
I could probably check if these Server Bundle secrets are already there with the proper name... if they are, then i skip the secret creation/override:

```
oc get secrets -n mas-mas-devops-manage | grep main-manage-d--sb
main-manage-d--sb0--asc--sn                             Opaque                                1      2m25s
main-manage-d--sb1--asc--sn                             Opaque                                1      2m25s
main-manage-d--sb2--asc--sn                             Opaque                                1      2m24s
main-manage-d--sb3--asc--sn                             Opaque                                1      2m24s
main-manage-d--sb4--asc--sn                             Opaque                                1      2m24s
```

if the above returns me something, then i skip it?

whitfiea (approved):

Issue #295, opened 2022-05-07
Deliver application-specific one-click playbooks


durera:
The naming convention isn't final yet, but we need to deliver the following playbooks
- [x] oneclick_add_assist
- [x] oneclick_add_hputilities
- [x] oneclick_add_iot @durera 
- [x] oneclick_add_manage @durera 
- [x] oneclick_add_monitor
- [x] oneclick_add_predict
- [x] oneclick_add_safety
- [x] oneclick_add_visualinspection

Each playbook should assume that the `oneclick_install_core` playbook has already been executed; it needs to install any dependencies for the application, install the application (at the most recent version supported by MAS 8.7.x), and configure the application in a workspace.

PR #777, opened 2023-04-18
[minor] oneclick_add_hputilities playbook

andrercm:
Addresses the remaining work for https://github.com/ibm-mas/ansible-devops/issues/295

- This extends one click playbook support to install Health & Predict Utilities application and its dependencies.
- Made some small updates to Predict one click playbook to keep consistency.
- Implemented logic to determine which version of AppConnect to install based off HPU channel, if AppConnect is chosen to be installed out of the oneclick playbook (AppConnect install is set to false by default until the security issues are resolved, customer would have to install at his own risk). There is still option to install AppConnect passing the specific channel/version parameters.


lokesh-sreedhara (dismissed):

andrercm:
HPU will be gone and deprecated in MAS 8.11 so this is not needed anymore.

PR #780, opened 2023-04-20
[minor] Support setting trustDefaultCAs in Suite

stonepd:
Set `settings.trustDefaultCAs` in the Suite CR as part of the suite_install role.
The `settings.trustDefaultCAs` flag is optional, set from the MAS_TRUST_DEFAULT_CAS environment variable if it is provided.


leo-davison (approved):

whitfiea (dismissed):
As this is adding features then it needs to be a [minor] and have the two changelogs updated to include this.

whitfiea:
@stonepd given that if we merge this today it won't actually cause the change we think it should (as no released MAS level has this function) then we should put this PR into Draft and then open again once we have a released product that supports this (and update the readme to indicate what level). Unless you think it is needed now?

ibm/mas_devops/roles/suite_install/README.md
```diff
@@ -24,6 +24,9 @@ Opitional fact, if not provided the role will use the default cluster subdomain
 ### mas_instance_id
 Defines the instance id to be used for MAS installation
 
+### mas_trust_default_cas
+Defines whether Default CAs are included in MAS trust stores. Can be set by the `MAS_TRUST_DEFAULT_CAS` environment variable
```
durera, line 28: Should state something like "only takes effect with MAS 8.11 or newer"

ibm/mas_devops/roles/suite_install/defaults/main.yml
```diff
@@ -16,6 +16,8 @@ mas_certificate_duration: "{{ lookup('env', 'MAS_CERTIFICATE_DURATION') | defaul
 mas_certificate_renew_before: "{{ lookup('env', 'MAS_CERTIFICATE_RENEW_BEFORE') | default('720h0m0s', true) }}"
 mas_manual_cert_mgmt: "{{ lookup('env', 'MAS_MANUAL_CERT_MGMT') | default(False, true) }}"
 
+mas_trust_default_cas: "{{ lookup('env', 'MAS_TRUST_DEFAULT_CAS') | bool }}"
```
durera, line 19: Should `| default(False, true)` before `| bool` otherwise we will pass "" to bool when env var not defined ... which probably evaluates to something, but it's not immediately clear to a reader what that result would be ... I would assume passing "" to bool = False at a guess .... but by passing to default first it's much clearer what the intent is.

ibm/mas_devops/roles/suite_install/templates/core_v1_suite.yml.j2
```diff
@@ -46,3 +46,7 @@ spec:
 {% if mas_img_pull_policy != '' %}
     imagePullPolicy: {{  mas_img_pull_policy  }}
 {% endif %}
+{% if mas_trust_default_cas is defined and mas_trust_default_cas != '' %}
```
durera, line 49: mas_trust_default_cas has been passed to `bool` it will never equal "" (empty string) .... I think.

IIRC earlier versions of MAS will accept this, so could just add it to all MAS instance settings.  Alternatively to keep with your intent:

`mas_trust_default_cas: "{{ lookup('env', 'MAS_TRUST_DEFAULT_CAS') }}" # don't cast to bool`

This would allow us to see that the var wasn't set, or was set to ""

then we can cast to bool in the template (or do it in elsewhere and save it as `mas_trust_default_cas_bool` or somesuch

ibm/mas_devops/roles/suite_install/README.md
stonepd: added "This only has an effect in IBM Maximo Application Suite version 8.11 and above."

ibm/mas_devops/roles/suite_install/defaults/main.yml
stonepd: Actually the requirement is to default this flag to true in line with current behaviour for the suite to trust defaultCAs, I'll add the explicit default to make it clear

ibm/mas_devops/roles/suite_install/templates/core_v1_suite.yml.j2
stonepd: Good point - I personally prefer only providing the suite flag when the env var is set, which avoids any confusion adding the flag to pre-8.11 Suite CRs. I'll do the boolean cast here in the template.

ibm/mas_devops/roles/suite_install/defaults/main.yml
stonepd: 2nd thoughts - I'll move the boolean cast to the Suite template so I can only add the field in the quite if the MAS_TRUST_DEFAULT_CAS env var is specifically defined.

stonepd: @durera - What should we do when the users set MAS_TRUST_DEFAULT_CAS to a value that is neither True or False?  Should the script fail, or provide the default value of "True"?
(currently export MAS_TRUST_DEFAULT_CAS=nonsense will set the suite trustDefaultCAs flag to false!)

stonepd:
I refactored handling of mas_trust_default_cas boolean so we don't include the trustDefaultCAs flag in the Suite CR when no value is provided.

I tested these changes locally in a test ansible playbook

durera (approved):

Issue #781, opened 2023-04-21
'dict object' has no attribute 'saslmechanism' in IoT due to kafka user error


whitfiea:
The configiration of IoT fails with `'dict object' has no attribute 'saslmechanism' ` due to the kafkacfg not being configured correctly (so the kafka binding fails to create the resource that contains the saslmechanism. 

The kafkacfg can't be validation as the ansible-devops role for kafka has somehow changed the masuser credentials in amq-streams but not updated the masuser crendentials that is given to the kafkacfg. It looks like running the ansible-devops role for kafka mulitple times changes the credentials so the role is not idemptotent.

andrercm:
I have a fix for it in a current branch I'm working on for event streams enhancements,

Issue #546, opened 2022-12-07
Kafka role assumes cluster monitoring is setup


whitfiea:
The kafka role was updated in PR https://github.com/ibm-mas/ansible-devops/pull/451 to include the grafana dashboards for kafka/zookeeper. But if you don't run the cluster_monitoring role beforehand (As you don't want the overheard) then the kafka role fails with:

```
TASK [ibm.mas_devops.kafka : Create Kafka monitoring dashboards] ***************************************************************************************************************************************************
failed: [localhost] (item=/Users/whitfiea/.ansible/collections/ansible_collections/ibm/mas_devops/roles/kafka/templates/dashboards/kafka-zookeeper.yml.j2) => {"ansible_loop_var": "item", "changed": false, "item": "/Users/whitfiea/.ansible/collections/ansible_collections/ibm/mas_devops/roles/kafka/templates/dashboards/kafka-zookeeper.yml.j2", "msg": "Failed to find exact match for integreatly.org/v1alpha1.GrafanaDashboard by [kind, name, singularName, shortNames]"}
failed: [localhost] (item=/Users/whitfiea/.ansible/collections/ansible_collections/ibm/mas_devops/roles/kafka/templates/dashboards/kafka.yml.j2) => {"ansible_loop_var": "item", "changed": false, "item": "/Users/whitfiea/.ansible/collections/ansible_collections/ibm/mas_devops/roles/kafka/templates/dashboards/kafka.yml.j2", "msg": "Failed to find exact match for integreatly.org/v1alpha1.GrafanaDashboard by [kind, name, singularName, shortNames]"}
failed: [localhost] (item=/Users/whitfiea/.ansible/collections/ansible_collections/ibm/mas_devops/roles/kafka/templates/dashboards/kafka-exporter.yml.j2) => {"ansible_loop_var": "item", "changed": false, "item": "/Users/whitfiea/.ansible/collections/ansible_collections/ibm/mas_devops/roles/kafka/templates/dashboards/kafka-exporter.yml.j2", "msg": "Failed to find exact match for integreatly.org/v1alpha1.GrafanaDashboard by [kind, name, singularName, shortNames]"}
```

The kafka role should only install these dashboards if the `integreatly.org/v1alpha1.GrafanaDashboard` CRD exists on the cluster

tcskill:
[...]

PR #782, opened 2023-04-24
[patch] check for exisitng user + supports integreatly

whitfiea:
Fixes https://github.com/ibm-mas/ansible-devops/issues/781 to make the kafka role idempotent. Now if you run the kafka role again on an existing environment then it will detect the maskafka-credentials secret in the kafka namespace, and if there is a password it will use that rather than create a new random password.

Also fixes https://github.com/ibm-mas/ansible-devops/issues/546 which now checks the cluster to see if it supports the api `integreatly.org/v1alpha1` and if it does then the grafanadashboard resources are installed.

This is the output from a rerun of the kafka role (you can see that it checked the cluster supported the `integreatly.org/v1alpha1` api, and that the kafka user was not modified) :

image

image

### DIFF
--- a/ibm/mas_devops/roles/kafka/tasks/provider/redhat/install.yml
+++ b/ibm/mas_devops/roles/kafka/tasks/provider/redhat/install.yml
@@ -13,7 +13,7 @@
 
 - name: Determine cluster capabilities
   set_fact:
-    supports_integreatly_org: "{{ 
+    supports_integreatly_org: "{{
       api_status is defined and
       api_status.apis is defined and
       api_status.apis['integreatly.org/v1alpha1'] is defined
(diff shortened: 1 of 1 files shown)


ibm/mas_devops/roles/kafka/tasks/provider/redhat/install.yml
```diff
@@ -4,7 +4,40 @@
...
+    supports_integreatly_org: "{{
+      api_status is defined and
+      api_status.apis is defined and
+      api_status.apis['integreatly.org/v1alpha1'] is defined
+    }}"
+
+
+# 3. Lookup any existing kafka user in the kafka namespace to use the same password
+# -----------------------------------------------------------------------------
+- name: "Lookup any exiting Kafka user Secret"
```
andrercm, line 25: minor typo in `exiting`

andrercm (dismissed):

andrercm (approved):

PR #783, opened 2023-04-24
[patch] Enhancements for Event Streams support

andrercm:


whitfiea (changes requested):
All looks good but can you make this a minor and add the changelog details as this is adding the ability to control if the topics should be created or not.

whitfiea (approved):

PR #796, opened 2023-05-04
[minor] Add support for tuning OCP ingress

cuddlyporcupine:
Add support for tuning the following OCP ingress parameters
- clientTimeout
- serverTimeout

### DIFF
--- a/ibm/mas_devops/roles/ingress/README.md
+++ b/ibm/mas_devops/roles/ingress/README.md
@@ -1,0 +1,38 @@
+# ingress
+This role will tune the Ingress operator in the target OCP cluster.
+
+## Role Variables
+### ingress_client_timeout
+specifies how long a connection is held open while waiting for a client response
+
+- Optional
+- Environment Variable: `INGRESS_CLIENT_TIMEOUT`
+- Default Value: `30s`
+
+### ingress_server_timeout
+specifies how long a connection is held open while waiting for a server response
+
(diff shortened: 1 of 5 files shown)


ibm/mas_devops/roles/ingress/README.md
```diff
@@ -0,0 +1,38 @@
+# ingress
+This role will tune the Ingress operator in the target OCP cluster.
+
+## Role Variables
+### ingress_client_timeout
+specifies how long a connection is held open while waiting for a client response
```
whitfiea, line 6: Capitalize "specifies"

whitfiea (changes requested):
As this is updating ocp can we change the name of the role from `ingress` to `ocp_ingress` to follow the convention we have. Names of varialbes should be OCP_INGRESS_xx as well.

As this is a minor and a new role then we need the changelogs updated https://github.com/ibm-mas/ansible-devops/blob/master/docs/changes.md and https://github.com/ibm-mas/ansible-devops/blob/master/ibm/mas_devops/README.md. 

The docs need the new role so you need to update https://github.com/ibm-mas/ansible-devops/blob/master/ibm/mas_devops/README.md with the role and https://github.com/ibm-mas/ansible-devops/blob/master/ibm/mas_devops/README.md

Can you also add a setence or two to the PR and the Readme to say why someone might want to change this.

PR #798, opened 2023-05-05
[minor] Add support for tuning OCP Ingress

cuddlyporcupine:
Add a new role to support tuning the following OCP ingress parameters. This role will tune the Ingress operator in the target OCP cluster. The following Ingress tuning parameters can be modified to avoid request failures due to timeout for long running requests.

- clientTimeout
- serverTimeout


whitfiea (approved):
LGTM

PR #809, opened 2023-05-12
[minor] Add OpenTelemetry to monitoring stack

xwgao:
This PR is for adding ansible role to support installation or uninstallation of OpenTelemetry operator which is necessary for enabling service monitoring of MAS & Apps with OpenTelemetry in on-prem envs.


xwgao:
This PR has been tested locally already, below are the local test results.

- Install OpenTelemetry operator

image

- Uninstall OpenTelemetry operator

image

Issue #801, opened 2023-05-09
Grafana is using the wrong storage class/access mode


durera:
Reviewing the monitoring stack we stand up, it appears that it's not been working for a while, depending on where the scheduler deploys the grafana pods.

We are currently using RWO storage class for Grafana: https://github.com/ibm-mas/ansible-devops/blob/master/ibm/mas_devops/roles/cluster_monitoring/tasks/install/determine-storage-classes.yml#L56-L66 and setting the access mode to `ReadWriteOnce` https://github.com/ibm-mas/ansible-devops/blob/master/ibm/mas_devops/roles/cluster_monitoring/templates/grafana.yml.j2#L24-L25

Both need to be changed, and we need to make sure that the CLI tekton pipeline is mapping the RWX storage class to the Grafana storage class parameter.

```
Events:
  Type     Reason       Age                    From            Message
  ----     ------       ----                   ----            -------
  Warning  FailedMount  2m14s (x454 over 17h)  kubelet         (combined from similar events): Unable to attach or mount volumes: unmounted volumes=[grafana-data], unattached volumes=[grafana-data grafana-provision-dashboards grafana-datasources kube-api-access-g2l5m grafana-config grafana-logs grafana-plugins grafana-provision-notifiers grafana-provision-plugins]: timed out waiting for the condition
  Warning  FailedMount  107s (x514 over 17h)   ibm/ibmc-block  Error while attaching the device pv pvc-4e01d5d0-a1f8-4e09-ab5a-c12e53f72245 cannot be attached to the node 10.183.59.178. Error: PV pvc-4e01d5d0-a1f8-4e09-ab5a-c12e53f72245 is already attached to another node 10.183.59.133 and there are active pods [grafana-deployment-69cccd89c-xwmx9] using that
```

Additionally the `cluster_monitoring` role should be updated such that it would fail if this happens .. it should be verifying the deployment is healthy and failing if it's not.

JonahLuckett:
Correction - #832 above is not linked to this issue

PR #814, opened 2023-05-16
[patch] Updating Grafana deployment strategy to Recreate

JonahLuckett:
### Description

- Instead of changing the grafana storage class as proposed in the original issue this change instead bumps the deployment strategy from `rollingUpdate` to `Recreate` which means that when grafana needs to be updated it will kill the first instance before moving onto the new one instead of having 2 open at once which is blocked by the storage class.
**Testing this in a current instance of this failure (fvtcore) resulting in correction meaning it is not only a preventative measure but also a proactive fix.**

- Adding Grafana CR readiness check to cluster_monitoring. Has a max 10 minute wait time which has proven to be more than enough time for a fresh install to get the grafana CR up and running.
**Output on fresh instance of the check happening:**
Screenshot 2023-05-25 at 14 45 24

### Related issues
- https://github.com/ibm-mas/ansible-devops/issues/801
- https://github.ibm.com/wiotp/tracker/issues/12548 (Internal)


JonahLuckett:
See cli equivalent: https://github.com/ibm-mas/ansible-devops/pull/814

whitfiea (approved):

PR #815, opened 2023-05-16
[minor] Add new SLS license upload mechanism and partly deprecated SLS bootstrap process

rawa-resul:
## Description

- Updated README
- Added new var `entitlement_file` - `SLS_ENTITLEMENT_FILE` for the new license upload process in SLS 3.7.0
- Removed unused workload scaling remnants
- Added checks to make sure old bootstrap and new upload mechanism don't conflict
- Enforcing license_id with license_file in bootstrap now (for up to 3.6.0)
- Added checks and verification for registrationKey bootstrap
- Tweaked sls_verify.yml tasks to provide smoother/faster experience

## Testing

Tested with SLS 3.6.0 and 3.7.0 using a combination of vars


whitfiea (changes requested):
@rawa-resul As we are deprecating some variables and adding a new one, this should be a minor change and have the two changelog updates

whitfiea (approved):

Issue #734, opened 2023-03-27
uds role should add proxy/cluster trustedCA to generated BAS configuration


aeisma:
The udscfg that the uds role generates contains certificate that the role copies from common_tasks/get_ingress_cert.yml.
This results in a reconciliation failure `BAS configuration was unable to be verified` in entitymanager-bascfg for an ingress certificate that is issued by a private authority with `[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get issuer certificate (_ssl.c:1131)`.

The reason is that the private root CA is not in the list of certificates copied into the udscfg by the uds role, as this root certificate is not in the ingress certificate.

I've manually added the root CA to the udscfg to be able to proceed, but the uds role could get and add the required root CA from the proxy/cluster configmap as follows:

```
oc get configmap -n openshift-config $(oc get proxy/cluster -o yaml -o custom-columns=":.spec.trustedCA.name" --no-headers) -o yaml -o custom-columns=":.data.ca-bundle\.crt" --no-headers
```

Note that proxy/cluster spec.trustedCA is empty when the cluster is using a self-signed ingress certificate, or an ingress certificate that is signed by a public authority.

andrercm:
Thanks @aeisma for you feedback, we'll try to have it resolve as part of another related issue we have opened around BAS/UDS certificates:
https://github.com/ibm-mas/ansible-devops/issues/721

JonahLuckett:
Closing with the merged PR above

PR #821, opened 2023-05-17
[patch] Add trustedCA to generated cert configuration

JonahLuckett:
## Description

Added a path on the get_ingress_cert which will use private root CA if provided instead of the signed cert path. Should have zero impact on the standard running of the roles - will only come up if a user has actively defined a private cert.

**Output with private cert:**
Screenshot 2023-05-19 at 14 37 16

**Output without a private cert:**
Screenshot 2023-05-19 at 14 39 40

### Related issues
- https://github.com/ibm-mas/ansible-devops/issues/734
- https://github.ibm.com/wiotp/tracker/issues/12429 _(Internal)_


whitfiea (approved):
LGTM

Issue #804, opened 2023-05-10
Tags are missing from the extras mirror manifest files


PeterNiblett:
The mirror manifest files produced by the mirror_case_prepare role contain image tags but those produced by  mirror_extras_prepare do not.   

Tags aren't strictly necessary for Airgap deployments since everything is pulled by digest rather than tag, however the oc mirror some local mirror registries seem to require the images to have a tag when things are pushed to them via the oc mirror command - an entry gets created in the registry but no image is saved, and the oc mirror command does not log any error when this happens.

Even if the local registry does accept images without tags, it is still helpful to users if the images are tagged with their versions.

PR #823, opened 2023-05-18
[patch] Add tags to the extras mirror manifest files

PeterNiblett:
This patch is for Airgap mirroring and is to resolve issue #804.  

Currently if you use the CLI to produce mirror manifest files, the ones that are  produced from CASE bundles (i.e. the mirror_case_prepare role) contain image tags but those like Mongo images,  produced by mirror_extras_prepare,  do not.

Some airgap mirror registries appear to require all images to be tagged. 

I have tested the mirror_extras_prepare role and have verified that the tags get included.


durera (changes requested):
Not all extras have tags ... e.g. 

```
extra_images:
  - name: cpopen/ibm-cpd-cloud-native-postgresql-operator-bundle
    registry: icr.io
    digest: sha256:3279e5dc5cccc23aafb4c48192b7fc4585a80fe232b65f4e1f562448ed7fc3c8

  - name: cpopen/ibm-cpd-cloud-native-postgresql-operator
    registry: icr.io
    digest: sha256:4ef97266b00aa94b697065434a46f43f064c1254fac4dab993998c5068f541d7
```

Every file in `ibm/mas_devops/roles/mirror_extras_prepare/vars` would need to have tags for every image, or this would break for those that do not.

PeterNiblett:
I have made the addtion of the image tag conditional on the tag being defined in the vars file, so it won't now break if we don't have tags there.  I have also added the tags to the db2 extras files.   The only file that is currently missing tags is the wd extras one.

PeterNiblett:
> Not all extras have tags ... e.g.
> 
> ```
> extra_images:
>   - name: cpopen/ibm-cpd-cloud-native-postgresql-operator-bundle
>     registry: icr.io
>     digest: sha256:3279e5dc5cccc23aafb4c48192b7fc4585a80fe232b65f4e1f562448ed7fc3c8
> 
>   - name: cpopen/ibm-cpd-cloud-native-postgresql-operator
>     registry: icr.io
>     digest: sha256:4ef97266b00aa94b697065434a46f43f064c1254fac4dab993998c5068f541d7
> ```
> 
> Every file in `ibm/mas_devops/roles/mirror_extras_prepare/vars` would need to have tags for every image, or this would break for those that do not.



> This patch is for Airgap mirroring and is to resolve issue #804.
> 
> Currently if you use the CLI to produce mirror manifest files, the ones that are produced from CASE bundles (i.e. the mirror_case_prepare role) contain image tags but those like Mongo images, produced by mirror_extras_prepare, do not.
> 
> Some airgap mirror registries appear to require all images to be tagged.
> 
> I have tested the mirror_extras_prepare role and have verified that the tags get included.

durera (approved):

Issue #835, opened 2023-05-25
Wrong digest for MongoDB in air-gap mirroring


witekwww:
There is wrong digest for MongoDB (quay.io/ibmmas/mongo) using v8-230414-amd64
Digest in the CASE points to version 4.2.23 and not to 4.4.21
This will result in mirroring the wrong image and air-gapped installation will fail due to missing image.

whitfiea:
@witekwww Can you use catalog [v8-230518-amd64](https://ibm-mas.github.io/cli/catalogs/v8-230518-amd64/) as the static catalog from last month won't have the recently updated mongo image like you have found.

whitfiea:
@witekwww After reviewing the digests we have I spotted the issue, and it is just like you say that the digest is incorrect. The reference PR should fix this once it is it in a release of mas_devops.

PR #837, opened 2023-05-25
[patch] Fix incorrect mongo 4.4.21 digest

whitfiea:
The wrong digest was added for the mongo 4.4.21 (it was copied from the 4.2.23 image digest). Without this change then the default MongoDb install on airgap can not work as the image is not mirrored.

This was missed in testing as we can't currently block quay.io https://github.com/ibm-mas/ansible-devops/blob/master/ibm/mas_devops/roles/ocp_simulate_disconnected_network/defaults/main.yml#L3

Resolves https://github.com/ibm-mas/ansible-devops/issues/835


caroazad (approved):
Double checked with quay.io and it's looking good to me.

durera (approved):

PR #838, opened 2023-05-25
[minor] Support May Catalog Update

just4li:


whitfiea (dismissed):
@just4li The PR needs a description. I am not sure we want to replace the existing catalog. We have had customers ask before why the static catalogs disappear some times, and this would be such a case. I understand that the reason is due to the Monitor 8.10.1 not being usable due to a digest issue, but the other apps are ok. I would think only if the catalog is completely useless then we would want to replace it rather than add this as just a newer catalog. A comment in the docs to say don't use catalog v8-230518-amd64 if you want to use Monitor would also help

Edit: I updated the wrong PR. I meant this one https://github.com/ibm-mas/cli/pull/317#pullrequestreview-1445570594

durera (approved):

PR #847, opened 2023-06-02
[patch] fix mas_app_channel when called via oneclick_upgrade playbook

andrercm:
Not my proudest PR, but this fixes the problem found by customer in this support ticket:
https://github.ibm.com/wiotp/tracker/issues/12685

The problem here is that we are setting fact on `mas_app_channel` here:
https://github.com/ibm-mas/ansible-devops/blob/master/ibm/mas_devops/roles/suite_app_upgrade/tasks/main.yml#L26

And if that happens once (in customer issue case, it happened when the playbook called the app upg role for the first time during Assist upgrade), it will always take precedence over the `mas_app_channel` var passed by the `oneclick_upgrade.yml` playbook, therefore it will not reset the value per role execution as originally intended by the playbook:
https://github.com/ibm-mas/ansible-devops/blob/master/ibm/mas_devops/playbooks/oneclick_upgrade.yml#L65

Customer is also saying that to avoid the error, they set the environment variable "SKIP_COMPATIBILITY_CHECK" to "true" to skip the compatibility check and run the Playbook "oneclick_upgrade" again. That overcomes the problem because if that flag is set, then the set fact that is originating the problem is not executed due the `when` condition not satisfied.

Before this fix, customer was expecting Manage to be upgraded to 8.6.x:

```
export UPGRADE_CORE=True
export UPGRADE_APPS=True
export MAS_CHANNEL=8.10.x
export MAS_APP_CHANNEL_MANAGE=8.6.x
```

but was getting this:

```
　[2023-04-20 14:04:05.112] TASK [ibm.mas_devops.suite_app_upgrade : manage : Check that upgrade is supported from current channel] *********************

　　　　　[2023-04-20 14:04:05.144] fatal: [localhost]: FAILED! => {

　　　　　[2023-04-20 14:04:05.144] "assertion": "app_sub_info.resources[0].spec.channel == upgrade_requirement[mas_app_channel]",

　　　　　[2023-04-20 14:04:05.144] "changed": false,

　　　　　[2023-04-20 14:04:05.144] "evaluated_to": false,

　　　　　[2023-04-20 14:04:05.144] "msg": "Unable to upgrade from 8.5.x to 8.7.x"
```

Because `mas_app_channel` was set as fact to `8.7.x` when the role was executed while attempting to upgrade Assist using the playbook.

If there's any smarter or better way to enforce use of vars being passed by the playbook even when set facts are performed by the inner roles, I'd be very happy to learn too, but that's the best I could do to have it properly working the way it was intended to work :)

That does not really change any behavior of the upgrade role, that just "adds" an internal var to be used to set the app channel allowing the original `mas_app_channel` to be overwritten by the playbook.


durera (commented):
Would a refactor to something like `mas_app_upgrade_target_channel` not make the role of this variable more obvious?

Also, there should be no need to null it at the end of the run either as we always set it to a new value at the start of the run (in main.yml L20-39).

andrercm:
> Would a refactor to something like `mas_app_upgrade_target_channel` not make the role of this variable more obvious?
> 
> Also, there should be no need to null it at the end of the run either as we always set it to a new value at the start of the run (in main.yml L20-39).

Good catch, i can clean that up now that we're using a new variable to control the target channel so no need to initialize it for next runs, and yes, i can rename the variable from `mas_app_channel_internal` to `mas_app_upgrade_target_channel`

andrercm:
@durera can you help re-review this?

whitfiea (approved):

PR #865, opened 2023-06-13
[patch] Fix db2u resource defaults

durera:
Previous bad merge resulted in a change to the default values for CPU & memory resource requests & limits.


andrercm (approved):
LGTM

Issue #795, opened 2023-05-03
suite_dns role: installing in an empty cluster has side effects on common-services install


jeancrg:
When I tried to run the role in an empty cluster that did not have cert-manager installed, the role failed, which is ok. However, later I tried the install of IBM Common services running its specific role and it failed

![image](https://user-images.githubusercontent.com/104657263/236009530-577ef0e6-e70e-4b82-8b60-f4c7254db252.png)

![image](https://user-images.githubusercontent.com/104657263/236009559-4325ef5c-297b-4f19-9e74-6f1ff3ea9fbb.png)

![image](https://user-images.githubusercontent.com/104657263/236009603-308d614e-419e-4d68-b6ac-7c62f29ccedb.png)

This solved the problem: oc delete apiservice [v1alpha1.acme.cis.ibm.com](http://v1alpha1.acme.cis.ibm.com/)

However needs to update the role to avoid this side effect to happen. Some validation or cleaning in case of failure.

whitfiea:
@andrercm maybe something to include when you look at https://github.com/ibm-mas/ansible-devops/issues/530?

andrercm:
I think we need to add an assert to check if cert-manager is installed rather than checking which cert-manager is installed (it's long past due since we dropped support for jetstack cert-manager)

andrercm:
@jeancrg this issue is fixed with the linked PR, will close this issue but we can reopen later if needed.

PR #870, opened 2023-06-16
[major] Remove old cert-manager related roles and adds new validations to `suite_dns` role

andrercm:
This PR will do the following:

- Removes pretty old roles designed to facilitate cert-manager resources migration during MAS 8.6 upgrade (doubt anyone making use of that still): `cert_manager_upgrade` and `cert_manager_upgrade_check` roles.
- Adds an assert to check if provided CIS CRN in fact corresponds to a valid CIS instance in the logged IBM Cloud account and prevents the `suite_dns` role from progressing otherwise. This will help avoid problems as raised in this issue: https://github.ibm.com/wiotp/tracker/issues/12729
- Adds an assert to check if ibm cert manager is installed and prevents the `suite_dns` role from progressing if it's not installed. This will help avoid problems as raised in this issue: https://github.com/ibm-mas/ansible-devops/issues/795 where the cis webhook and api services are installed while cert manager might not be, which will cause the api service to hang.


durera (changes requested):
> Removes pretty old roles designed to facilitate cert-manager resources migration during MAS 8.6 upgrade

Needs to be marked as a **major** change (v14.0) as we are removing capability.

whitfiea (approved):
LGTM

ibm/mas_devops/roles/suite_dns/tasks/providers/cis/main.yml
```diff
@@ -1,4 +1,13 @@
 ---
+- name: "cis : lookup cis instances with provided 'cis_crn' value"
+  shell: ibmcloud cis instances | grep {{ cis_crn }} | awk '{print $2}'
```
rhocheck, line 3: the CIS plugn is not available on all platforms, e.g. Linux aarch64. why not do
```
ibmcloud resources | grep {{ cis_crn }} | awk '{print $2}'
```

PR #872, opened 2023-06-19
[patch] Add core and apps' digests for 8.9 June release

sanju7216:
Since Airgap v2 support is not backported to MAS 8.9, we ll need to add digests for Core and apps as long as we support 8.9


caroazad (approved):
looking good to me

durera (approved):

andrercm (approved):

PR #884, opened 2023-06-26
[patch] Add HPU for 810 airgap

caroazad:
- add hpu digest image-map file 
- Don't check for Trust Store manager for HPU because HPU doesn't use it.


andrercm (approved):

andrercm (approved):
LGTM

PR #890, opened 2023-06-28
[minor] Add support for CloudPak for Data 4.6.6

andrercm:
This PR introduces support for [CPD 4.6.6 and its services](https://github.ibm.com/wiotp/tracker/issues/12488).
It requires a catalog source from `cpd466` [branch](https://github.ibm.com/maximoappsuite/ibm-maximo-operator-catalog/tree/cpd466/ibm-operator-catalog/2023-06-28-curated-amd64/cloud-native-postgresql) to be installed as it contains change needed to install Postgres (WD dependencies) up to 1.18.3 version only (which is the version that works with WD 4.6.5 / CPD 4.6.6), if Postgres 1.18.4+ is installed then WD 4.6.5 does not install properly. This is the specific [commit](https://github.ibm.com/maximoappsuite/ibm-maximo-operator-catalog/commit/4b7f7cfd4619799cf4fc4bc2c5a9241fe360552f) that we need in the catalog source.

Also removed the `cpd_cr_lookup.resources[0].status.wosStatus == "Failed"` condition from `openscale wait task` because sometimes during openscale install, CR sets status to `failed` but it self heals a couple of minutes later.


andrercm:
All work is done code wise, the problem though is that WD 4.6.6 does not successfully install.

andrercm:
Started the work to support CPD 4.6.6, all CPD services can be installed/upgraded just fine with exception of Watson Discovery. PR with the code changes to support it are currently in [draft](https://github.com/ibm-mas/ansible-devops/pull/890) until we get to understand what's wrong with Watson Discovery, the install gets stuck at the middle of the install, we have same issue as reported here:
https://ibm-watson-iot.slack.com/archives/CJQ323MMM/p1687771973000059

I tried to install previous version WD 4.6.3 using v8-cpd466-amd64 catalog but that also does not work, so it makes me think that the issue is coming from one of the new operand versions for Watson Discovery or its dependencies like Minio or Elasticsearch operators, as WD 4.6.3 install is working fine with v8-master-amd64 as it's been daily installed with our FVTs.

sanju7216 (approved):

PR #918, opened 2023-07-17
[patch] Make ephemeral storage size configurable for SNO

Aymen-Ben-S:
Tested end to end with this PR content in https://dashboard.masdev.wiotp.sl.hursley.ibm.com/tests/fvtsno

image


alequint (approved):

ibm/mas_devops/roles/ocp_provision/templates/ipi/install-config.yaml.j2
```diff
@@ -13,6 +13,8 @@ controlPlane:
   platform:
     {{ ipi_platform }}:
       type: {{ ipi_controlplane_type }}
+      rootVolume:
+        size: {{ ipi_storage_size }}
```
alequint, line 17: we are saying here that all IPI installation will use ephemeral storage of 300... shouldn't we, instead, surround it with a conditional to undertand if we are using SNO - similar with [this](https://github.ibm.com/maximoappsuite/ansible-fvt/pull/241/files#r7767476). Criteria could be the same used in CLI to identify we are provisioning SNO, Let's discuss.

alequint (changes requested):

ibm/mas_devops/roles/ocp_provision/templates/ipi/install-config.yaml.j2
Aymen-Ben-S: @alequint how about we leverage this variable that I found used in cli?

alequint: @Aymen-Ben-S pls look at https://github.ibm.com/maximoappsuite/ansible-fvt/pull/241#discussion_r7775965

alequint (changes requested):

ibm/mas_devops/roles/ocp_provision/templates/ipi/install-config.yaml.j2
durera: Please don't introduce sno mode specific _anything_ ... ipi_storage_size is something that is valid sno or not sno ... sno is still just openshift.

` if ipi_storage_size is defined and not ""` is sufficient, there's no reason to limit support for this to sno only.  If someone wants to customize this field .. SNO or non-SNO they can set IPI_STORAGE_SIZE

ibm/mas_devops/roles/ocp_provision/defaults/main.yml
```diff
@@ -84,6 +84,7 @@ ipi_compute_replicas: "{{ lookup('env', 'IPI_COMPUTE_REPLICAS') | default('3', t
 # For now we can default the type to the best fit for AWS as that's all we support atm
 ipi_controlplane_type: "{{ lookup('env', 'IPI_CONTROLPLANE_TYPE') | default('m5.4xlarge', true) }}"
 ipi_controlplane_replicas: "{{ lookup('env', 'IPI_CONTROLPLANE_REPLICAS') | default('3', true) }}"
+ipi_storage_size: "{{ lookup('env', 'IPI_STORAGE_SIZE') | default('300', true) }}"
```
durera, line 87: By defaulting this to 300 we are setting this value for 99% of customers who will not customize this ... is that really the intent here?  What value would this normally be if not explicitly set?

Suggest it's better to not a default for this, unless someone sets this we continue to function as we did before, by not setting `rootVolume.size` .. also suggest calling it `ipi_rootvolume_size` so it's more reflective of it's use.  Also, this needs to be added to the documentation in the role's `README.md`

durera (changes requested):

alequint (dismissed):

alequint:
It looks good to me now @Aymen-Ben-S

alequint (approved):

andrercm (approved):

PR #930, opened 2023-07-27
[minor] Add mongo ce monitoring

cuddlyporcupine:
Add support for MongoDB CE monitoring for operator versions 0.7.8 and higher
- Add Prometheus block to MongoDBCommunity CR
- Add metrics endpoint secret
- Add service monitor
- Add Grafana dashboard

image


durera (dismissed):
Can you address the merge conflicts, also can you advise on the relationship between this role and the cluster monitoring role ... this looks like it's all required, so the cluster monitoring role must now be executed before running the mongo role?

If this is the case, this should be documented in the role's README.md, and there should be a related update to the pipelines in CLI so that cluster-monitoring is set as a dependency for the mongo step.

cuddlyporcupine:
@durera , the merge conflicts have been addressed and I've updated the mongodb role README to mention the dependency on the cluster_monitoring role

ibm/mas_devops/roles/mongodb/README.md
```diff
@@ -20,6 +20,7 @@ Also, you need to have AWS user credentials configured via `aws configure` comma
 
 To run the `docdb_secret_rotate` MONGODB_ACTION when the provider is `aws` you must have already installed the [Mongo Shell](https://www.mongodb.com/docs/mongodb-shell/install/).
 
+To run this role with provider `community` you must first run the [cluster_monitoring role](https://ibm-mas.github.io/ansible-devops/roles/cluster_monitoring/). The grafana dashboard and prometheus monitoring resources depend on the `cluster_monitoring` role.
```
whitfiea, line 23: Can we change this statement to say something like "This role will install a GrafanaDashboard used for monitoring the MongoDB instance whe the provider is `community` and you have run the [cluster_monitoring role](https://ibm-mas.github.io/ansible-devops/roles/cluster_monitoring/) previously. If you did not run the [cluster_monitoring role](https://ibm-mas.github.io/ansible-devops/roles/cluster_monitoring/) then the GrafanaDashboard won't be installed."

This is based on the change to guard against installing the grafanadashboard later on.

ibm/mas_devops/roles/mongodb/tasks/providers/community/install.yml
```diff
@@ -247,6 +267,48 @@
...
+# -----------------------------------------------------------------------------
+- name: "community : install : Check for existing Grafana dashboard"
+  kubernetes.core.k8s_info:
+    api_version: integreatly.org/v1alpha1
+    kind: GrafanaDashboard
+    name: mongodb-overview
+    namespace: "{{ mongodb_namespace }}"
+  register: mongoce_grafana_dashboard_info
+  when:
+    - mongodb_ce_version is version('0.7.8', '>=')
```
whitfiea, line 301: Can we add a guard against this task like we do for kafka https://github.com/ibm-mas/ansible-devops/blob/3ebec8c56a69959001ee6388d28f5e376996cd52/ibm/mas_devops/roles/kafka/tasks/provider/redhat/install.yml#L178 that will check if the integreatly.org api is installed or not. This will allow the mongodb role to continue to run if cluster_monitoring was not run.

Note: you would need to add a lookup for the api like here https://github.com/ibm-mas/ansible-devops/blob/3ebec8c56a69959001ee6388d28f5e376996cd52/ibm/mas_devops/roles/kafka/tasks/provider/redhat/install.yml#L41

whitfiea (changes requested):
@cuddlyporcupine I made some comments about changing the code so we can make the cluster_monitoring role optional. There are a number of customers that don't want cluster_monitoring due to the extra resource cost, and this change would force them to have it as most rely on the community mongodb. The change would make it optional as the grafanadashboards is the only thing that would fail as the servicemonitor CRD is already installed in the base OCP.

ibm/mas_devops/roles/mongodb/README.md
cuddlyporcupine: @whitfiea above changes have been made, currently building

ibm/mas_devops/roles/mongodb/tasks/providers/community/install.yml
cuddlyporcupine: @whitfiea requested changes have been made, currently building

cuddlyporcupine:
@whitfiea requested updates have been made

whitfiea (approved):

PR #933, opened 2023-07-28
[patch] Fix CPD 4.6.6 change logs

andrercm:


durera (commented):
This is no longer required .. we are using the GHE releases page as the change log now .. it's automatically generated, and doesnt need any manual action.

PR #939, opened 2023-08-01
[patch] Add info around route53 workaround

andrercm:
Include documentation referenced to a workaround that might be needed in cert-manager operator in case Route 53 DNSs are not resolved properly.

This **workaround is not automated**,  because it requires the `cert-manager-controller` pod to be stopped/disabled, thus we want to avoid making this behavior as standard approach in the automation.


ibm/mas_devops/roles/suite_dns/README.md
```diff
@@ -202,6 +202,11 @@ Role Variables - AWS Route 53
 To run this role successfully you must have already installed the [AWS CLI](https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html).
 Also, you need to have AWS user credentials configured via `aws configure` command or simply export `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` environment variables with your corresponding AWS username credentials prior running this role.
 
+**Note:** In some cases, the Route 53 zone may not resolve the certificate challenges generated by IBM Certificate Manager, which could cause a problem while issuing the public certificates via Let's Encrypt. In this case, a manual workaround might be needed in cert-manager-controller pod to enable recursive nameservers.
```
durera, line 205: Can you format this like:

```
!!! note
    In some cases, the Route 53 zone may not resolve the certificate challenges generated by IBM Certificate Manager, which could cause a problem while issuing the public certificates via Let's Encrypt. In this case, a manual workaround might be needed in cert-manager-controller pod to enable recursive nameservers.
```

This will format this as an admonition (the coloured boxes)

durera (changes requested):

ibm/mas_devops/roles/suite_dns/README.md
andrercm: Done.

durera (approved):

PR #951, opened 2023-08-09
[patch] adding new role to perform rollback on mas suite

sekharcvalluri:
A new role (suite_rollback) has been added to perform rollback operation on mas suite. Currently it works only for core, however, it will be updated soon to have rollback feature for apps like manage etc.


rawa-resul (approved):

ibm/mas_devops/roles/suite_rollback/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+- **Required**
+- Environment Variable: `MAS_CORE_VERSION`
+- Default: None
+
+### mas_rollback_dryrun
+When set to `true` will ensure that the role only preforms rollback validation checks and does not make any changes to the target installation.
+
+- Optional
+- Environment Variable: `MAS_ROLEBACK_DRYRUN`
```
whitfiea, line 46: typo, should be MAS_ROLLBACK_DRYRUN

ibm/mas_devops/roles/suite_rollback/defaults/main.yml
```diff
@@ -0,0 +1,10 @@
+---
+mas_rollback_dryrun: "{{ lookup('env', 'MAS_ROLLBACK_DRYRUN') | default('False', True) | bool }}"
+
+mas_core_version: "{{ lookup('env', 'MAS_CORE_VERSION') }}"
+rollback_mas_core: "{{ lookup('env', 'ROLLBACK_MAS_CORE') | default('True', True) | bool }}"
+verify_core_version: "{{ lookup('env', 'VERIFY_CORE_VERSION') | default('False', True) | bool }}"
+mas_instance_id: "{{ lookup('env', 'MAS_INSTANCE_ID') }}"
+mas_namespace: "mas-{{ mas_instance_id }}-core"
+
+skip_compatibility_check: "{{ lookup('env', 'SKIP_COMPATIBILITY_CHECK') | default('False', True) | bool }}"
```
whitfiea, line 10: Should SKIP_COMPATIBILITY_CHECK be in the docs?

ibm/mas_devops/roles/suite_rollback/meta/main.yml
```diff
@@ -0,0 +1,23 @@
+galaxy_info:
+  author: André Marcelino (@andrercm)
+  description: Check an existing IBM Maximo Application Suite v8.11.0 installation is ready to rollback
```
whitfiea, line 3: We should remove 8.11.0 from this description. As I assume we can rollback 8.12 etc?

ibm/mas_devops/roles/suite_rollback/README.md
```diff
@@ -0,0 +1,78 @@
+suite_rollback
+===============================================================================
+This role validates if a given MAS installation is ready for the core platform to be rolled back to a specific MAS core version, and (as long as dry run mode is not enabled) will execute the rollback.
```
whitfiea, line 3: Is there a level that this won't work at? i.e. MAS 8.9.? Is the user allowed to rollback minor levels as well as patch levels? Might need some guidance as to what is expected i.e. Can I rollback 8.11.1 to 8.9.4?

whitfiea (changes requested):

ibm/mas_devops/roles/suite_rollback/README.md
sekharcvalluri: Corrected

ibm/mas_devops/roles/suite_rollback/defaults/main.yml
sekharcvalluri: SKIP_COMPATIBILITY_CHECK environment is meant only for development. It should be always 'False' by default, however, in dev mode, it is set to 'True'. If the user tries to rollback to any version which is not in supported list, it should fail with a proper message to the user. However, in the dev, we accept rollback from one pre built version to another on the same base version though it is not listed under supported versions. For ex: 8.11.0-pre.dev to 8.11.0-pre.stable.

ibm/mas_devops/roles/suite_rollback/README.md
sekharcvalluri: Updated the Readme with more details.

ibm/mas_devops/roles/suite_rollback/defaults/main.yml
whitfiea: @sekharcvalluri Can you add a comment to this file just above skip_compatibility_check so that we know it is for dev reasons only and is not expected to be changed by an end user?

sekharcvalluri: Done

whitfiea (approved):

Issue #934, opened 2023-07-28
Cipher compatibility issue with default IngressController configuration in FIPS mode


terenceq:
## No cipher suite in common with Semeru runtime and default IngressController configuration

This problem can be seen when a Java client (for example a connection from the Manage all or mea containers) attempts a connection to the `auth` route. An example is a connection from a Java client using IBM Semeru to the URL:

```
https://<auth route>/oidc/endpoint/MaximoAppSuite/token"
```

The [default tlsSecurityPolicy](https://docs.openshift.com/container-platform/4.12/security/tls-security-profiles.html#tls-profiles-ingress-configuring_tls-security-profiles) that is configured for the `IngressController` does not contain a cipher in common with IBM Java Semeru when running in FIPS mode. A proposed solution is to configure the `tlsSecurityProfile` in a similar fashion as what is done for the  `APIServer`. For example:

```yaml
 tlsSecurityProfile:
   custom:
     ciphers:
       - TLS_AES_128_GCM_SHA256
       - TLS_AES_256_GCM_SHA384
       - TLS_CHACHA20_POLY1305_SHA256
       - ECDHE-ECDSA-AES128-GCM-SHA256
       - ECDHE-RSA-AES128-GCM-SHA256
       - ECDHE-ECDSA-AES256-GCM-SHA384
       - ECDHE-RSA-AES256-GCM-SHA384
       - ECDHE-ECDSA-CHACHA20-POLY1305
       - ECDHE-RSA-CHACHA20-POLY1305
       - DHE-RSA-AES128-GCM-SHA256
       - DHE-RSA-AES256-GCM-SHA384
       - ECDHE-RSA-AES128-SHA256
       - ECDHE-RSA-AES128-SHA
       - ECDHE-RSA-AES256-SHA
     minTLSVersion: VersionTLS12
   type: Custom
```

The above configuration should be added to the `IngressController` resource (usually named default)

An interesting observation is that out of the box on Fyre when FIPS was enabled the IngressController did not seem to be enforcing the default profile of `Intermediate`. However, as soon as the `Intermediate` policy was set Semeru clients running in FIPS mode could no longer connect to the auth route.

Again for reference see the docs here:

https://docs.openshift.com/container-platform/4.12/security/tls-security-profiles.html#tls-profiles-ingress-configuring_tls-security-profiles

terenceq:
[...]

PR #963, opened 2023-08-16
[major] FIPS configuration for IngressController, removes OCP_APISERVER_UPDATE_CIPHERS option

terenceq:
## IngressController tlsSecurityProfile change required for FIPS

When in FIPS mode the `IngressController` requires a change to its `tlsSecurityProfile` - this change is similar to what has been done for the `APIServer`. This is required because MAS Semeru Java clients must have at least one cipher in common when in FIPS mode. By default this is not the case. 

Until IBM Java Semeru adds more cipher support in FIPS mode these type of configuration changes will be necessary in FIPS mode.

This PR also creates a more generic task to update both the `APIServer` and `IngressController` in one task.

https://github.com/ibm-mas/ansible-devops/issues/934

**Breaking change**: Remove `OCP_APISERVER_UPDATE_CIPHERS` option in `ocp_config` and replaces it with `OCP_UPDATE_CIPHERS_FOR_SEMERU`


ibm/mas_devops/roles/ocp_config/defaults/main.yml
```diff
@@ -2,7 +2,7 @@
 # Configuration Settings
 # -----------------------------------------------------------------------------
 ocp_ingress_update_timeouts: "{{ lookup('env', 'OCP_INGRESS_UPDATE_TIMEOUTS') | default('False', true) | bool }}"
-ocp_apiserver_update_ciphers: "{{ lookup('env', 'OCP_APISERVER_UPDATE_CIPHERS') | default('False', true) | bool }}"
+ocp_update_ciphers_for_semeru: "{{ lookup('env', 'OCP_UPDATE_CIPHERS_FOR_SEMERU') | default('False', true) | bool }}"
```
whitfiea, line 5: As the `OCP_APISERVER_UPDATE_CIPHERS` is already in a release we need to either continue to support it (but deprecate it and also add OCP_UPDATE_CIPHERS_FOR_SEMERU that sets the same thing) or just make this a major change.

terenceq: I've updated the PR title and pushed a major commit. 

Also tested one more time with a Fyre cluster configured for FIPS.

whitfiea (approved):

PR #969, opened 2023-08-18
[minor] Add support to upgrade OCS/ODF

caroazad:
Add support to upgrade channel for OCS/ODF
- ocs to odf (test passed)
Tested from ocs on ocp 4.10 to odf on ocp 4.11 ( I had to upgrade the ocp versions before the upgrades)
- odf to odf (test passed)
Tested from odf on ocp 4.11 to odf on ocp 4.12 ( I had to upgrade the ocp versions before the upgrades)
Also run some other tests for if there is no storage operator in the cluster and if it was already up to date (all tests passed)

Discovered a bug in common services upgrade where if it is not in the cluster, the pipeline fails. I fixed it by checking the subscription and displaying the message that there is nothing to upgrade for common services. 

Added environment var OCS_ACTION to control it. Default is install. Set to upgrade to upgrade the storage if needed depending on ocp version

After the operators are upgraded, the storage cluster version is also updated to match the ocp version 

* I found out the hard way that the cli changes does not show in the cluster after provisioning so that is why I have all those commits. I was making changes thinking that it wasn't working. I don't know if this is a bug but it was very frustrating.


ibm/mas_devops/roles/ocs/defaults/main.yml
```diff
@@ -1,3 +1,6 @@
 ---
 # The device path needed for the local volume used by the local storage operator
 lso_device_path: "{{ lookup('env', 'LSO_DEVICE_PATH') | default('/dev/vdb', True) }}"
+
+# ocs action
+ocs_action: "{{ lookup('env', 'OCS_ACTION') | default('install', True) }}"
```
whitfiea, line 6: Can we add this new paramter to the documentation?

whitfiea (changes requested):
Can you make this a "minor" rather than "patch" as we are adding function. Also could you expand on the "test passed" part in the PR i.e. what version did you test to and from?

ibm/mas_devops/roles/ocs/defaults/main.yml
caroazad: Done

whitfiea (approved):

PR #966, opened 2023-08-17
[patch] multiple quick winner fixes

andrercm:
This PR fixes the following issues:

- https://github.com/ibm-mas/ansible-devops/issues/923
  - added option to receive the values from a var instead of just using the default value for each app
- https://github.com/ibm-mas/ansible-devops/issues/955
  - added `no_log: true` in some `set_facts` and masked passw and tokens from debugs  
- https://github.com/ibm-mas/ansible-devops/issues/883
  - added some more info around the storage class vars in MVI playbook
- https://github.com/ibm-mas/ansible-devops/issues/625
  - added the same input validation we have in mas cli for `mas_instance_id` and `mas_workspace_id`
- https://github.com/ibm-mas/ansible-devops/issues/896
  - added a statement to kafka role to describe mas license does not include amq streams entitlement

Also added the fix for `ocp_config` to avoid the issues with redhat catalog.

PR #970, opened 2023-08-18
ocp_config the ocp_operatorhub_disable_redhat_sources default value typo

conradkao:
The default value should be either True or False


andrercm:
Dup of https://github.com/ibm-mas/ansible-devops/pull/966

PR #979, opened 2023-08-22
[minor] Optional Watson Studio bind to Health

alequint:
Following https://github.com/ibm-mas/ansible-devops/pull/957 - two main points about this PR:

- Update manage playbook offering possibility to install cp4d and studio to be bound with Health
- Modified he binding parameter (where we would pass a free string, expecting `system`) to a boolean flag. Reason for that is because `cp4d_service`role, today, when installing Studio, does not give us an option to create cfg in a different level inside MAS, only `system` is applicable (check [here](https://github.com/ibm-mas/ansible-devops/blob/master/ibm/mas_devops/roles/cp4d_service/tasks/gencfg/gencfg-wsl.yml#L145)); same way, would not make any sense offer a choice at this point (if we want to bind, `True`, it will be a `system` binding)


ibm/mas_devops/roles/suite_app_config/README.md
```diff
@@ -129,6 +129,13 @@ URL to access WML service (same as Cloud Pak for Data URL).
 
 Role Variables - Manage Workspace
 -------------------------------------------------------------------------------
+
+### mas_appws_bindings_health_wsl_flag
+Optional. Boolean value indicating if Watson Studio must be bound to Manage. It is expected a sytem level WatsonStudioCfg applied in the cluster.
```
andrercm, line 134: minor typo `sytem`

andrercm (changes requested):
just a minor typo found, other than that i think it's looking ok.

ibm/mas_devops/roles/suite_app_config/README.md
alequint: done

andrercm (approved):
LGTM

PR #983, opened 2023-08-25
[minor] Use image setting from catalog for mongodb role

racree:
Update mongodb role to use image settings from the catalog and mirror extras roles.
With this change the mongodb role requires the catalog_tag.  
It uses the catalog_tag to determine the mongo version and then uses the mongo version to determine the image settings.


ibm/mas_devops/roles/mongodb/tasks/providers/community/install.yml
```diff
@@ -1,6 +1,51 @@
 ---
 
-# 1. Debug Properties
+# 1. Check for undefined properties that do not have a default
+# -----------------------------------------------------------------------------
+- name: "Fail if catalog_tag is not provided"
+  assert:
+    that: catalog_tag is defined and catalog_tag != ""
```
whitfiea, line 7: Can we call it mas_catalog_version as that is what it is rather than introducing a different name

whitfiea (changes requested):
Now that the MONGODB_CE_VERSION is no longer being used we will need to update the docs here https://github.com/ibm-mas/ansible-devops/blob/eb996518a5b5f5b8e7e533328b6b2ef454b334a3/ibm/mas_devops/roles/mongodb/README.md#mongodb_ce_version and add the new requirement for the MAS_CATALOG_VERSION to be set. We should state what happens if you change this and what it means for mongo.
 
 Also can you add what testing you have done with this change to the PR.

racree:
install1

racree:
log1

racree:
version1

PR #994, opened 2023-09-01
[major] Support Strimzi as default Kafka provider

andrercm:
Adding support for Strimzi as default Kafka provider in `kafka` role.
Also adding validation to skip and avoid installing Strimzi if AMQ Streams is already installed in the defaulted namespaces (lookup either in `amq-streams` or in the defined `kafka_namespace`) as Strimzi and AMQ uses the same apis.


andrercm:
Test is in progress

ibm/mas_devops/playbooks/awsrootca.pem
```diff
@@ -0,0 +1,20 @@
...
+93FcXmn/6pUCyziKrlA4b9v7LWIbxcceVOF34GfID5yHI9Y/QCB/IIDEgEw+OyQm
+jgSubJrIqg0CAwEAAaNCMEAwDwYDVR0TAQH/BAUwAwEB/zAOBgNVHQ8BAf8EBAMC
+AYYwHQYDVR0OBBYEFIQYzIU07LwMlJQuCFmcx7IQTgoIMA0GCSqGSIb3DQEBCwUA
+A4IBAQCY8jdaQZChGsV2USggNiMOruYou6r4lK5IpDB/G/wkjUu0yKGX9rbxenDI
+U5PMCCjjmCXPI6T53iHTfIUJrU6adTrCC2qJeHZERxhlbI1Bjjt/msv0tadQ1wUs
+N+gDS63pYaACbvXy8MWy7Vu33PqUXHeeE6V/Uq2V8viTO96LXFvKWlJbYK8U90vv
+o/ufQJVtMVT8QtPHRh8jrdkPSHCa2XV4cdFyQzR1bldZwgJcJmApzyMZFo6IQ6XU
+5MsI+yMRQ+hDKXJioaldXgjUkK642M4UwtBV8ob2xJNDd2ZhwLnoQdeXeGADbkpy
+rqXRfboQnoZsG4q5WTP468SQvvG5
+-----END CERTIFICATE-----
```
whitfiea, line 20: What uses this pem file?

whitfiea (changes requested):
Changes look good. I would make this a [major] as the behaviour is changing and someone rerunning this role after upgradeing the ansible collection would get a failure (as amq-streams would have been installed and now we we would try to install strimzi as default). There was also a question on the pem file.

ibm/mas_devops/roles/kafka/tasks/provider/redhat/check-amq-exists.yml
```diff
@@ -0,0 +1,40 @@
+---
```
durera, line 1: Think needs to be more generic and work both ways.

- If install strimzi check for redhat in the cluster
- If install redhat check for strimzi in the cluster

Use `operatorcondition` as the check ... e.g. ansible equivalent of:

```
oc get operatorcondition -l operators.coreos.com/amq-streams.amq-streams --all-namespaces
NAMESPACE     NAME                  AGE
amq-streams   amqstreams.v2.4.0-3   17h
```

And the equivalent for Strimzi.

I'm pretty sure there's example of this in the collection somewhere already too.

ibm/mas_devops/playbooks/awsrootca.pem
andrercm: thanks for spotting that, this file is created during msk provisioning. I'll remove that.

ibm/mas_devops/roles/kafka/tasks/provider/redhat/check-amq-exists.yml
andrercm: That's good, thanks! I'll make these changes.
The only thing that concerns me is that both current and this approach uses namespace scoped lookups, so if users that changed the default namespaces while installing their kafka providers, we won't be able to determine if they're installed or not.

andrercm: In any case, i made the proposed changes so now it is generic enough to works both ways.

If installing strimzi and amq is already present in default namespace:
```
TASK [ibm.mas_devops.kafka : Assert: Stop Strimzi install if Red Hat AMQ Streams is already installed] ********************************************************************************************************************************************************
fatal: [localhost]: FAILED! => {
    "assertion": "not lookup_kafka_provider_exists",
    "changed": false,
    "evaluated_to": false,
    "msg": [
        "An existing Red Hat AMQ Streams installation was identified in this cluster under namespace: amq-streams",
        "Strimzi cannot be installed because it conflicts with Red Hat AMQ Streams service APIs. Please, uninstall Red Hat AMQ Streams before installing Strimzi.",
        "If uninstalling Red Hat AMQ Streams is not an option, then set 'kafka_provider: redhat' to continue using Red Hat AMQ Streams as this is the current installed Kafka provider."
    ]
}
```

If installing amq and strimzi is already present in default namespace:
```
TASK [ibm.mas_devops.kafka : Assert: Stop Red Hat AMQ Streams install if Strimzi is already installed] ********************************************************************************************************************************************************
fatal: [localhost]: FAILED! => {
    "assertion": "not lookup_kafka_provider_exists",
    "changed": false,
    "evaluated_to": false,
    "msg": [
        "An existing Strimzi installation was identified in this cluster under namespace: strimzi",
        "Red Hat AMQ Streams cannot be installed because it conflicts with Strimzi service APIs. Please, uninstall Strimzi before installing Red Hat AMQ Streams.",
        "If uninstalling Strimzi is not an option, then set 'kafka_provider: strimzi' to continue using Strimzi as this is the current installed Kafka provider."
    ]
}
```

whitfiea (approved):

durera (approved):

PR #999, opened 2023-09-06
[minor] Automation for cleaning up mirror registry

caroazad:
Add support to remove the registry from the cluster so that it can be recreated and started at fresh.


ibm/mas_devops/roles/registry/README.md
```diff
@@ -1,7 +1,7 @@
 registry
 =======
 
-Create a Docker Registry running on RedHat OpenShift cluster.  The registry will be backed by persistant storage, and accessible via either a clusterIP or loadbalancer service.
+Create a Docker Registry running on RedHat OpenShift cluster.  The registry will be backed by persistant storage, and accessible via either a clusterIP or loadbalancer service. This role can also be used to cleanup the registry or remove the registry so that it can be recreated and have a clean start.
```
whitfiea, line 4: Can we add any implications for the tear-down action? I assume that it means if a pod restarts and the registry is no longer present then the image can't be pulled.

whitfiea (changes requested):

ibm/mas_devops/roles/registry/README.md
caroazad: @whitfiea I added some more explanation to the docs. Please check and let me know if there is something to change or reword, etc. Thank you!

whitfiea (approved):
Thanks for the excellent write up

PR #1004, opened 2023-09-12
[minor] suite_app_verify role

alequint:
To be used as the first task of `fvt-manage` pipeline. It is similar to `suite-verify` for MAS core, but it just check Workspace CR based on parameters... it is the same checking made in the end of `suite-app-config` therefore that piece of code has been moved to this role and then reused there) - we can consider use these "verify" tasks also before we start upgrade process to guarantee things are all set for the applications we want to bring to new versions.

Tested in internal pipelines successfully (evidences to be posted soon)

Related PRs:
- https://github.com/ibm-mas/cli/pull/455
- https://github.com/ibm-mas/ansible-devops/pull/1004
- https://github.ibm.com/maximoappsuite/ansible-fvt/pull/281
- https://github.ibm.com/maximoappsuite/fvt-ibm-mas-manage/pull/111


ibm/mas_devops/roles/suite_app_config/tasks/main.yml
```diff
@@ -101,21 +101,5 @@
 # 8. Wait for application workspace to be ready
 # -----------------------------------------------------------------------------
 - name: "Wait for application workspace to be ready ({{ mas_app_cfg_delay }}s delay)"
```
andrercm, line 103: What if we change the step name to something else than `- name: "Wait for application workspace to be ready ({{ mas_app_cfg_delay }}s delay)"` ? because the `suite_app_verify` will call the same step again, so we end up having two steps with same name?

andrercm (changes requested):

ibm/mas_devops/roles/suite_app_config/tasks/main.yml
alequint: Agree

andrercm (approved):

ibm/mas_devops/roles/suite_app_config/tasks/main.yml
alequint: Done

Issue #962, opened 2023-08-16
oneclick_core unnecessary validation for SLS_LICENSE_ID and SLS_LICENSE_FILE 


witekwww:
oneclick_core returns error if SLS_LICENSE_ID or SLS_LICENSE_FILE is not provided. However, both variables are depraceted with SLS 3.7.0.
Setting SLS_LICENSE_ID variable causes sls role to crash with "When using entitlement_file (SLS 3.7.0 or higher) do not set bootstrap.license_id."

Solution: update oneclick_core pre_tasks - line 31 and 32

andrercm:
thanks for the feedback @witekwww 
can you please share the exported variables used in this case while running `oneclick_core` playbook?
this will help us debug the problem.

witekwww:
@andrercm - sure thing.
when using variables as described in SLS role (https://ibm-mas.github.io/ansible-devops/roles/sls/)
export SLS_ENTITLEMENT_FILE=/mnt/home/license.dat
and running ansible-playbook ibm.mas_devops.oneclick_core
i'm getting error like this:
![image](https://github.com/ibm-mas/ansible-devops/assets/118370754/89fc45b7-3511-45a0-bfa9-e2b1d76a8e76)

however if i remove line 31 and 32 from https://github.com/ibm-mas/ansible-devops/blob/master/ibm/mas_devops/playbooks/oneclick_core.yml it goes through ok

the solution is to remove verification of SLS_LICENSE_ID var and change verification of SLS_LICENSE_FILE var to SLS_ENTITLEMENT_FILE

andrercm:
So there are some changes we'll have to do to better validate these fields now that SLS 3.7 was introduced.

i think your initial problem was that you might be exporting both `SLS_LICENSE_FILE` (this is likely being set as you said you have to remove line 32 from the [playbook](https://github.com/ibm-mas/ansible-devops/blob/master/ibm/mas_devops/playbooks/oneclick_core.yml#L32)) and `SLS_ENTITLEMENT_FILE` variables? And currently there's a validation that blocks you from using both as it will be one or the other depending on the SLS version you are using, that's why it is stated:
```
"When using entitlement_file (SLS 3.7.0 or higher) do not set bootstrap.license_id."
```

And also, you are right, `SLS_LICENSE_ID` is still being validated in oneclick_core, which needs fixing as this is not used anymore in SLS 3.7+
As stated in the documentation it will only be used in previous versions of SLS:
```
[...]

PR #1014, opened 2023-09-19
[patch] Fix SLS version specific differences

rawa-resul:


andrercm:
@rawa-resul can we get the `oneclick_core` fixed also to not required `SLS_LICENSE_ID` anymore? I think we could remove all the required SLS properties from the playbook itself and hand it over to the sls role to deal with? 
https://github.com/ibm-mas/ansible-devops/blob/master/ibm/mas_devops/playbooks/oneclick_core.yml#L31

andrercm (changes requested):
Looks good to me, I'd just leverage this to also fix the oneclick_core playbook to adjust the updated checks/assertions

durera (approved):

rawa-resul:
> @rawa-resul can we get the `oneclick_core` fixed also to not required `SLS_LICENSE_ID` anymore? I think we could remove all the required SLS properties from the playbook itself and hand it over to the sls role to deal with? https://github.com/ibm-mas/ansible-devops/blob/master/ibm/mas_devops/playbooks/oneclick_core.yml#L31
> 
> We even have an issue for that: #962

good point - will remove that check in one_click and let the sls role handle rest otherwise it will get messy.

PR #1022, opened 2023-09-25
[patch] set UDS_ACTION to  install_suds in FIPS mode

conradkao:
When `OCP_FIPS_ENABLED` is true, we will install the dummy UDS by setting UDS_ACTION to install_suds since the UDS is not FIPS compliant


andrercm (approved):
lgtm

PR #1038, opened 2023-10-06
[patch] Fix support for kafka/ocs in mas update  

racree:
while testing mas update cli I found these 2 code changes to be necessary.  
There's an update for kafka provider path and an update to ocs.

### DIFF
--- a/ibm/mas_devops/roles/kafka/tasks/main.yaml
+++ b/ibm/mas_devops/roles/kafka/tasks/main.yaml
@@ -6,4 +6,4 @@
       - "Kafka Action .................... {{ kafka_action }}"
 
 - name: Provision or Deprovision Kafka Instance
-  include_tasks: ../roles/tasks/provider/{{ kafka_provider }}/{{ kafka_action }}.yml
+  include_tasks: ../roles/kafka/tasks/provider/{{ kafka_provider }}/{{ kafka_action }}.yml
--- a/ibm/mas_devops/roles/kafka/tasks/main.yaml
+++ b/ibm/mas_devops/roles/kafka/tasks/main.yaml
@@ -6,4 +6,4 @@
       - "Kafka Action .................... {{ kafka_action }}"
 
 - name: Provision or Deprovision Kafka Instance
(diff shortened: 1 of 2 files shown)


ibm/mas_devops/roles/kafka/tasks/main.yaml
```diff
@@ -6,4 +6,4 @@
       - "Kafka Action .................... {{ kafka_action }}"
 
 - name: Provision or Deprovision Kafka Instance
-  include_tasks: tasks/provider/{{ kafka_provider }}/{{ kafka_action }}.yml
+  include_tasks: ../roles/kafka/tasks/provider/{{ kafka_provider }}/{{ kafka_action }}.yml
```
whitfiea, line 9: This shouldn't be needed. What was being run when you saw this problem? It sounds like the role path was not set correctly?

whitfiea (changes requested):

ibm/mas_devops/roles/kafka/tasks/main.yaml
racree: I was testing "mas update" and it consistently fails with an invalid path that includes the directory "playbooks". Thus this change prepends "../roles/kafka/".

So there must be an problem with the role path for update in the cli?

whitfiea: @racree Was this running inside the cli docker container or just running `mas update` directly on your machine? We don't seem to need this change in other places so I suspect something might be changing the path before this.

racree: I ran the following steps inside the cli docker container.  
Steps:

1. mas install
2. installed kafka using "ROLE_NAME=kafka ansible-playbook ibm.mas_devops.run_role"
3. mas update

mas update consistently failed with a file not found.

whitfiea: the failure happens in the tekton pipeline, so don't think this environmental

whitfiea (approved):

PR #1043, opened 2023-10-11
[minor] Add Mongo v6 support

andrercm:


terenceq (approved):

durera (changes requested):
As discussed in Slack

durera (approved):

PR #1049, opened 2023-10-16
[minor] Add DRO role to aid migration from UDS

yuvraj-vansure:
This epic is basically a requirement to continue MAS's use of metrics and metering. UDS is getting sunset in approximately 2Q 2024, and we need a replacement technology. IBM Software Central has stepped up to the plate and will be extending the scope of the RHM operator to do this. This will be released as the "Data Reporter Operator", or DRO.

https://jsw.ibm.com/browse/MASCORE-16

DRO will be supported on the following MAS versions:
- MAS 8.10.6 +
- MAS 8.11.2 +
- MAS 9.0 +

The default provider will still contunue to be UDS but the option is there to use DRO instead if on a MAS level that supports it


ibm/mas_devops/playbooks/oneclick_core.yml
```diff
@@ -16,6 +16,7 @@
     # Workspace Configuration
     mas_workspace_name: "{{ lookup('env', 'MAS_WORKSPACE_NAME') | default('MAS Development', true) }}"
     mas_workspace_id: "{{ lookup('env', 'MAS_WORKSPACE_ID') | default('masdev', true) }}"
+    bas_action: "{{ lookup('env', 'BAS_ACTION') | default('UDS', True) }}"
```
lokesh-sreedhara, line 19: @yuvraj-vansure can u bas_action to always be upper case? Incase users set BAS_ACTION to lower case 'uds' or 'dro'

for example
`{{ value | upper }}`

lokesh-sreedhara (changes requested):

ibm/mas_devops/roles/dro/tasks/gencfg/main.yml
```diff
@@ -0,0 +1,38 @@
+---
+# 1. Set dro_tls_crt
+# -----------------------------------------------------------------------------
+- name: "gencfg : Set dro_tls_crt fact for BasCfg when using existing dro"
+  when:
+    - dro_tls_crt is not defined or dro_tls_crt == ""
+    - dro_tls_crt_local_file_path is defined and dro_tls_crt_local_file_path != ""
```
whitfiea, line 7: Is dro_tls_crt_local_file_path defined anywhere else?

ibm/mas_devops/roles/dro/tasks/uninstall/main.yml
```diff
@@ -0,0 +1,129 @@
+---
+# 1. Delete CR, CSV and Subscription
+# -----------------------------------------------------------------------------
+# Delete the MarketplaceConfig
+- name: "uninstall : Delete the MarketplaceConfig"
```
whitfiea, line 5: Do we want to delete the marketplace config? what if the user had redhat marketplace already installed?

ibm/mas_devops/roles/dro/tasks/uninstall/main.yml
```diff
@@ -0,0 +1,129 @@
...
+    state: absent
+    api_version: operators.coreos.com/v1alpha1
+    kind: Subscription
+    name: redhat-marketplace-operator
+    namespace: redhat-marketplace
+    wait: yes
+    wait_timeout: 300
+
+# Delete CRs
+- name: "uninstall : Delete the MarketplaceConfig"
```
whitfiea, line 83: seems to be a duplicate of the first task

docs/playbooks/oneclick-core.md
```diff
@@ -48,9 +48,14 @@ The other values can be left at their defaults.  Finally, click **Generate** and
 - `MAS_CONFIG_DIR` Directory where generated config files will be saved (you may also provide pre-generated config files here)
 - `SLS_LICENSE_ID` The license ID must match the license file available in `SLS_LICENSE_FILE`
 - `SLS_LICENSE_FILE` The path to the location of the license file.
-- `UDS_CONTACT_EMAIL` Defines the email for person to contact for UDS
-- `UDS_CONTACT_FIRSTNAME` Defines the first name of the person to contact for UDS
-- `UDS_CONTACT_LASTNAME` Defines the last name of the person to contact for UDS
+- `BAS_ACTION` Defines UDS or DRO operator to install. Default: UDS [Values: DRO, UDS]
```
whitfiea, line 51: Not sure this is an "action" but more of a "provider" or "type". We don't want to get this confussed with the install, uninstall actions.

ibm/mas_devops/roles/dro/README.md
```diff
@@ -0,0 +1,103 @@
+dro
```
whitfiea, line 1: As this is a new role then we need to update https://github.com/ibm-mas/ansible-devops/blob/master/mkdocs.yml to contain it.

Also if DRO will only work with certain MAS versions then it should be state in the docs.

whitfiea (changes requested):

ibm/mas_devops/roles/dro/tasks/uninstall/main.yml
yuvraj-vansure: MarketplaceConfig is created and managed by ibm-metrics-operator. There are no specific labels/annotations to identify if the MarketplaceConfig is installed by user before MAS installation or is created by MAS. 
I'll try to add some kind of check to identify existing MarketplaceConfig.

ibm/mas_devops/roles/dro/tasks/uninstall/main.yml
yuvraj-vansure: removed duplicate tasks

ibm/mas_devops/playbooks/oneclick_core.yml
yuvraj-vansure: updated task to make var always upper case

lokesh-sreedhara (dismissed):

ibm/mas_devops/roles/dro/README.md
whitfiea: @yuvraj-vansure @lokesh-sreedhara this question is still outstanding: 
> Also if DRO will only work with certain MAS versions then it should be state in the docs.

I believe we are backporting changes into certain levels of MAS, so if a user was to try to install DRO but had version 8.10.1 of MAS then I don't believe it would work?

lokesh-sreedhara: @whitfiea jus updated the readme with the supported MAS versions.

whitfiea (approved):

PR #1050, opened 2023-10-16
[patch] Do not validate storage class for mongo update

racree:
Storage class isn't set in the update pipeline.  The mongo role just needs to not do anything with storage classes in the action=update path, update doesn't need to change the storage.

Therefore the update action should not call tasks/determine-storage-classes.yml


racree:
ran "mas install" specifying catalog from Aug 29, 2023

Screenshot 2023-10-16 at 1 54 58 PM

MongoDB successfully installed
Screenshot 2023-10-16 at 1 57 07 PM

Then ran "mas update" specifying catalog from Oct 4, 2023
Screenshot 2023-10-16 at 1 55 17 PM

Update pipeline successfully ran

Screenshot 2023-10-16 at 1 58 55 PM

andrercm (approved):
makes sense to me

rawa-resul (approved):

ibm/mas_devops/roles/mongodb/tasks/main.yml
```diff
@@ -12,6 +12,7 @@
   when:
     - mongodb_provider == "community"
     - mongodb_action != "none"
+    - mongodb_action != "update"
```
whitfiea, line 15: There is no such action as update for mongo, we just call install again don't we?

whitfiea (changes requested):

ibm/mas_devops/roles/mongodb/tasks/main.yml
racree: changed "update" to "upgrade" which is consistent with other roles and created an "upgrade" action.

whitfiea (changes requested):
@racree I am not sure I understand how this fixes the problem stated, which was to not validate the storage class for mongo?

For the upgrade flow, it will call install, and if the install needs to upgrade (say from 0.7.8 to 0.7.9) then it will install this new CR https://github.com/ibm-mas/ansible-devops/blob/master/ibm/mas_devops/roles/mongodb/templates/community/0.7.9/cr.yml.j2#L75C35-L75C56 which has the mongodb storage class. So either the update needs to know the storage class or it has to read what the current storageclass is and set that in the new CR.

racree:
@whitfiea Please correct me if I misunderstand the following flow...

When the mongodb role is called with mongodb_action == upgrade...

First mongodb/tasks/main.yml is called and with this PR change will skip calling tasks/determine-storage-classes.yml and proceed to call tasks/providers/community/upgrade.yml.

Then tasks/providers/community/upgrade.yml directly calls tasks/providers/community/install.yml without calling back into main.yml.

Therefore with this PR, tasks/determine-storage-classes.yml is not called when mongodb_action == upgrade.

racree:
@whitfiea Does this PR need to be written differently?  My original understanding was... when Mongo is already installed and running, we do not need to validate the storage class because we should not be changing the storage class.   But based on your comment above, you are expecting something different?

racree:
@whitfiea I think I understand what you are saying... it is not enough to skip calling "tasks/determine-storage-classes" because the cr that gets applied during the update requires mongodb_storage_class to be defined.

racree:
fixed in https://github.com/ibm-mas/cli/pull/674

PR #1083, opened 2023-10-31
[minor] Support v8-231031-amd64 Catalog update

sanju7216:
Support for october 30 catalog - https://github.ibm.com/maximoappsuite/ibm-maximo-operator-catalog/pull/322

### DIFF
--- a/ibm/mas_devops/common_vars/casebundles/v8-231031-amd64.yml
+++ b/ibm/mas_devops/common_vars/casebundles/v8-231031-amd64.yml
@@ -1,0 +1,87 @@
+---
+# Case bundle configuration for IBM Maximo Operator Catalog v230926
+# -----------------------------------------------------------------------------
+# In the future this won't be necessary as we'll be able to mirror from the
+# catalog itself, but not everything in the catalog supports this yet (including MAS)
+# so we need to use the CASE bundle mirror process still.
+
+catalog_digest: sha256:TBD
+
+# Dependencies
+# -----------------------------------------------------------------------------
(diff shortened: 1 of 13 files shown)


ibm/mas_devops/common_vars/casebundles/v8-231031-amd64.yml
```diff
@@ -0,0 +1,87 @@
...
+  8.11.x: 8.9.0   # No Update
+  8.9.x: 8.7.1    # No Update
+
+# Extra Images for UDS
+# ------------------------------------------------------------------------------
+uds_extras_version: 1.4.0
+
+# Extra Images for Mongo
+# ------------------------------------------------------------------------------
+mongo_extras_version: 4.4.21
```
andrercm, line 79: Can we have something as this going forward?

```
mongo_extras_version_default: 4.4.21
mongo_extras_version: "{{ lookup('env', 'MONGODB_VERSION') | default(mongo_extras_version_default, True) }}"
```

We need to have enough flexibility to set `MONGODB_VERSION` in our fvt pipelines so that we can have mongodb version being rotated in our daily fvt runs.

andrercm (changes requested):

ibm/mas_devops/common_vars/casebundles/v8-231031-amd64.yml
sanju7216: Do you want me to make that change in this PR?? Don't want to mix that into catalog support PR. We can create a new PR for that once this is out. what do you think?

durera (approved):

PR #1084, opened 2023-10-31
[patch] Fix lots of timing issues

durera:
In many places people seem to have thought that `delay` is measured in minutes, leading to 1 second delays instead of 60 second delays.


sanju7216 (approved):
lgtm

stonepd (approved):

PR #1090, opened 2023-11-03
[patch] Restore default value for MAS_CONFIG_SCOPE

durera:
This default was accidentally removed in #1021, creating a breaking change which this update fixes.

Also:
- Updates docsite page width to match the expanded witdh used in CLI repo
- Significant updates to the Manage Playbook and Gen JDBC Cfg role documentation
- Moves role execution out of pre_tasks into roles


ibm/mas_devops/roles/gencfg_jdbc/README.md
```diff
@@ -3,51 +3,97 @@ gencfg_jdbc
...
 
+
+Role Variables - MAS Configuration
+-------------------------------------------------------------------------------
+### mas_config_scope
+Configure whether to generate a binding suitable for System, Workspace, Application, or Workspace-Application use within MAS (`system`, `ws`, `app`, or `wsapp`).
+
+- **Required**
+- Environment Variable: `MAS_CONFIG_SCOPE`
+- Default: None
```
istrate, line 60: Default here should be `system`

durera: There's no default in the role:

```
# MAS instance configuration
mas_config_scope: "{{ lookup('env', 'MAS_CONFIG_SCOPE')}}"
```

The default is only applied at the playbook level:
```
    # Manage components and server bundle configurations
    mas_config_scope: "{{ lookup('env', 'MAS_CONFIG_SCOPE') | default('system', true) }}" # Supported values are "system", "ws", "app", or "wsapp"
    mas_appws_bindings_jdbc: "{{ lookup('env', 'MAS_APPWS_BINDINGS_JDBC') | default('system', true) }}"
```

PR #1094, opened 2023-11-06
[minor] Enable ipv6 testing internally for Fyre

xinyuwu1:
This PR added the ability to deploy an internal fyre cluster with IPv6 enabled, and added suite CR properties for a future release
Added a flag enable_ipv6 to indicate whether it's for an ipv6-enabled cluster
Added code necessary to enable ipv6 on fyre (RPT only)
If enable_ipv6 is true, suite CR has the PreferDualStack and ["IPv6", "IPv4"] properties to run tests in this configuration.

Tested with pfvt for both enable_ipv6 true on FYRE and false (by default) on ibmcloud.  Provisioned and pipeline started running correctly


joaopauloksn (changes requested):

ibm/mas_devops/roles/ocp_config/tasks/main.yml
```diff
@@ -25,3 +25,7 @@
 # -----------------------------------------------------------------------------
 - include_tasks: update-operatorhub.yml
   when: ocp_operatorhub_disable_redhat_sources
+
+# 5. Enable dual stack for ipv6 test
```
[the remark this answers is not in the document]
xinyuwu1: I'd like to keep it there in case we still need it in the future. Currently it seems like this step is one-time deal only, if you run it manually once you don't need to do it again. But in case things change in the future I'd like to come back to this.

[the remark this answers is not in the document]
joaopauloksn: Well it is not a good practice to keep code commented-out. If you think it is really needed please add more context to it.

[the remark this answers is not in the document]
xinyuwu1: removed

joaopauloksn (dismissed):

ibm/mas_devops/roles/ocp_provision/defaults/main.yml
```diff
@@ -88,3 +89,7 @@ ipi_config_dir: "{{ ipi_dir }}/config/{{ cluster_name }}"
 
 ocp_installer_dir: "{{ ipi_dir }}/installer/{{ ocp_version }}"
 ocp_installer_exe: "{{ ipi_dir }}/installer/{{ ocp_version }}/openshift-install"
+
+# Dualstack required?
+ipv6_test: "{{ lookup('env', 'IPV6_TEST') | default('false', true) }}"
+fyre_root_password: "{{ lookup('env', 'FYRE_ROOT_PASSWORD') | default('false', true) }}"
```
whitfiea, line 95: I don't think we use `fyre_root_password` anywhere do we?

ibm/mas_devops/roles/ocp_login/tasks/login-fyre.yml
```diff
@@ -37,3 +37,47 @@
...
+    name: "cluster"
+  when:
+    - ipv6_test=="true"
+  register: _network
+
+- name: "fyre : Debug network"
+  debug:
+    var: _network
+
+# 7.For IPv6 testing, check network type
```
whitfiea, line 55: We should remove the commented out code

ibm/mas_devops/roles/ocp_config/defaults/main.yml
```diff
@@ -14,3 +14,6 @@ ocp_ingress_server_timeout: "{{ lookup('env', 'OCP_INGRESS_SERVER_TIMEOUT') | de
 # Ingress Controller Settings
 # -----------------------------------------------------------------------------
 ocp_operatorhub_disable_redhat_sources: "{{ lookup('env', 'OCP_OPERATORHUB_DISABLE_REDHAT_SOURCES') | default('False', true) | bool }}"
+
+# Dualstack required?
+ipv6_test: "{{ lookup('env', 'IPV6_TEST') | default('false', true) }}"
```
whitfiea, line 19: I don't think this default is used in this role.

ibm/mas_devops/roles/suite_app_install/defaults/main.yml
```diff
@@ -39,3 +39,6 @@ mas_app_settings_monitor_deployment_size: "{{ lookup('env', 'MAS_APP_SETTINGS_MO
 
 # Custom Labels
 custom_labels: "{{ lookup('env', 'CUSTOM_LABELS') | default(None, true) | string | ibm.mas_devops.string2dict() }}"
+
+# Dualstack required?
+ipv6_test: "{{ lookup('env', 'IPV6_TEST') | default('false', true) }}"
```
whitfiea, line 44: I don't think this var is used in this role

ibm/mas_devops/roles/suite_install/defaults/main.yml
```diff
@@ -116,3 +116,6 @@ mas_img_pull_policy: "{{ lookup('env', 'MAS_IMG_PULL_POLICY') }}"
 # Custom Labels
 # -----------------------------------------------------------------------------
 custom_labels: "{{ lookup('env', 'CUSTOM_LABELS') | default(None, true) | string | ibm.mas_devops.string2dict() }}"
+
+# Dualstack required?
+ipv6_test: "{{ lookup('env', 'IPV6_TEST') | default('false', true) }}"
```
whitfiea, line 121: Similar to the same var name in ocp_provision, we should call this enable_dualstack or something similar. It should also be added to the docs but with a statement that this is not supported yet.

ibm/mas_devops/roles/ocp_login/defaults/main.yml
```diff
@@ -28,3 +28,6 @@ supported_cluster_types:
   - aws
 
 k8s_auth_host: "{{ lookup('env', 'K8S_AUTH_HOST') | default('', True) }}"
+
+# Dualstack required?
+ipv6_test: "{{ lookup('env', 'IPV6_TEST') | default('false', true) }}"
```
whitfiea, line 33: I am not sure that ipv6_test is the best name. For this role it would be enable_dualstack ? Also this var needs to be added to the docs https://github.com/ibm-mas/ansible-devops/blob/master/ibm/mas_devops/roles/ocp_login/README.md?plain=1#L56

whitfiea (changes requested):

ibm/mas_devops/roles/ocp_provision/defaults/main.yml
xinyuwu1: It doesn't look like fyre_root_password is used. Let me remove that and retest

ibm/mas_devops/roles/ocp_login/defaults/main.yml
xinyuwu1: I'm using ipv6_test because that's the flag's name in devIT's api to request a fyre cluster of this flavor.  Is that a good reason enough to keep the name ipv6_test? @whitfiea

xinyuwu1: I think I'll call it enable_ipv6 then. Might not be dual stack in the future

ibm/mas_devops/roles/ocp_provision/README.md
```diff
@@ -23,6 +23,11 @@ Required.  Specify the version of OCP to install.  The exact format of this will
 - Environment Variable: `OCP_VERSION`
 - Default Value: None
 
+### enable_ipv6
```
whitfiea, line 26: this should be under the fyre section on line 225, and made clear it is fyre only

ibm/mas_devops/roles/suite_install/defaults/main.yml
```diff
@@ -116,3 +116,6 @@ mas_img_pull_policy: "{{ lookup('env', 'MAS_IMG_PULL_POLICY') }}"
 # Custom Labels
 # -----------------------------------------------------------------------------
 custom_labels: "{{ lookup('env', 'CUSTOM_LABELS') | default(None, true) | string | ibm.mas_devops.string2dict() }}"
+
+# Enable IPv6?
+enable_ipv6: "{{ lookup('env', 'ENABLE_IPV6') | default('false', true) }}"
```
whitfiea, line 121: this needs an entry in https://github.com/ibm-mas/ansible-devops/blob/4de7f6f21eb2244872f919a736d9a9750cba2429/ibm/mas_devops/roles/suite_install/README.md and needs to state what it does, and that it is currently only supported in internal devit fyre clusters.

The variable should also have the `| bool` filter at the end as it a boolean. Any checks can then check it as a boolean rather than a "true" string.

ibm/mas_devops/roles/suite_install/tasks/main.yml
```diff
@@ -169,6 +169,7 @@
       - "MAS ICR cpopen content ........ {{ mas_icr_cpopen }}"
       - "MAS Cluster Issuer ............ {{ mas_cluster_issuer }}"
       - "MAS Subscription Upgrade ...... {{ mas_upgrade_strategy }}"
+      - "IPv6 test...................... {{ enable_ipv6 }}"
```
whitfiea, line 172: `IPv6 test` should be `IPv6 Enabled`

ibm/mas_devops/roles/ocp_provision/defaults/main.yml
```diff
@@ -89,3 +89,6 @@ ipi_config_dir: "{{ ipi_dir }}/config/{{ cluster_name }}"
 
 ocp_installer_dir: "{{ ipi_dir }}/installer/{{ ocp_version }}"
 ocp_installer_exe: "{{ ipi_dir }}/installer/{{ ocp_version }}/openshift-install"
+
+# Enable IPv6?
+enable_ipv6: "{{ lookup('env', 'ENABLE_IPV6') | default('false', true) }}"
```
whitfiea, line 94: This should be filtered as a bool

whitfiea (changes requested):

PR #1096, opened 2023-11-10
[patch] Fix missing mirroring for Amlen operator image

durera:
After improving our airgap testing environments to block quay.io we disovered that we have a gap in the image mirroring for IoT; the Amlen operator image and bundle image are not being mirrored along with the rest of the IoT content.  This update remedies that by introducing a new amlen extras configuration.

This configuration has been applied to all catalogs since the July 25th catalog update.  Catalog versions before July are already unsupported due to end of life for OCP 4.10 in September.


IanBoden (commented):
looks good

PR #1102, opened 2023-11-13
[patch] mongodb doc updates for v5 and v6

terenceq:
### More details related to MongoDB 5 and 6

Adding a bit more context around the support of MongoDB 5 and 6. Specifically call out the following:

- MongoDB 4.4 will reach end of life soon
- Always take backups before any upgrade
- Specifically which versions of MongoDB 5 and 6 the automation can configure
- If MongoDB hosted by third party or on prem consult applicable documentation


andrercm (approved):
LGTM

PR #1105, opened 2023-11-14
[patch] extend the size of a page in returned dns entries

Aymen-Ben-S:
This will adjust the max number of dns entries as we've already exceeded 2k. 3500 is the CIS limit, we will add pagination if the call turned out to be unreliable.


alequint (approved):

alequint:
We need a closer look at this code and refactor it a little bit. In the mean time, I think we can keep the same design temporarily to make address certificate issues where redirect is not updated properly.

PR #1116, opened 2023-11-27
[patch] Predict w/ Watson OpenScale - DataMart ID Support

karol-czarnecki:
https://jsw.ibm.com/browse/MASHP-549
https://jsw.ibm.com/browse/MASHP-240


ibm/mas_devops/roles/suite_app_config/vars/defaultspecs/predict.yml
```diff
@@ -5,6 +5,9 @@ mas_appws_spec:
     jdbc: "{{ mas_appws_bindings_jdbc | default( 'system' , true) }}"
     watsonstudio: system
   settings:
+    mat:
+      datamartid: "{{ cpd_wos_datamart_id }}"
+      install: true
```
durera, line 10: Do we want this to always be true like this, for all versions of predict, even when cpd_wos_datamart_id == ""?

ibm/mas_devops/roles/suite_app_config/vars/defaultspecs/predict.yml
```diff
@@ -5,6 +5,9 @@ mas_appws_spec:
     jdbc: "{{ mas_appws_bindings_jdbc | default( 'system' , true) }}"
     watsonstudio: system
   settings:
+    mat:
+      datamartid: "{{ cpd_wos_datamart_id }}"
```
durera, line 9: What if cpd_wos_datamart_id == "", from the doc update this is something related to OpenScale, but that's an optional dependency I thought, so shouldn't be only be conditionally adding this if OpenScale is used?

Also, in a fully auotmated install, how does one obtain this ID?  Does it come from OpenScale, if so we need to be able to pass this value into the role via the install pipeline when/if OpenScale is selected for install.

durera (changes requested):

ibm/mas_devops/roles/suite_app_config/vars/defaultspecs/predict.yml
karol-czarnecki: Hi @durera
for install MAT we have 2 option: true/false
```
if MAT install true - then we need to set CPD_INSTALL_PLATFORM true/false
if MAT install false - then end

if CPD_INSTALL_PLATFORM true - then we need to set CPD_INSTALL_OPENSCALE true
if CPD_INSTALL_OPENSCALE true - then we need to set CPD_WOS_DATAMART_ID
if CPD_WOS_DATAMART_ID true - then end

if CPD_INSTALL_PLATFORM false - then we need provide following variables:
- CPD_WML_URL
- CPD_ADMIN_USERNAME
- CPD_ADMIN_PASSWORD
- CPD_WOS_DATAMART_ID
then end
```

ibm/mas_devops/roles/suite_app_config/vars/defaultspecs/predict.yml
karol-czarnecki: Is my previous comment answer this first part of the question?

I can see you did something similar for project_id here: https://github.com/ibm-mas/ansible-devops/blob/master/ibm/mas_devops/roles/suite_app_config/tasks/determine-watson-studio-id.yml

I will work on logic for CPD_WOS_DATAMART_ID - do you want me to follow your approach?

Related to automation install. I have found that we can use: 
- Creating a service instance for Watson OpenScale programmatically
- create datamart ids using Watson OpenScale Python SDK’s

ibm/mas_devops/roles/suite_app_config/vars/defaultspecs/predict.yml
durera: So, MAT is related to whether OpenScale is being used or not?  This is an optional dependency for Predict in the install logic so when it's not installed shouldn't this all be gated something like:

```
{% if cpd_wos_datamart_id is defined and cpd_wos_datamart_id  != "" %}
    mat:
      datamartid: "{{ cpd_wos_datamart_id }}"
{% endif %}
```

Otherwise we will set this to true for every install, whether OpenScale is set up or not.

karol-czarnecki:
Hi @durera 
I've change logic like you suggested. Could you please review? 

Regards

durera:
Closed due to old age/abandoned

PR #1121, opened 2023-11-30
[minor] Added failure handling for Maxinst setup script in suite_db2_setup_for_manage role

sachbalag:
The purpose of this PR is to make the db2 setup role more resilient so it fails immediately when there is an issue with the database, rather than failing silently. This fix handles the rc return value from the setupdb.sh script and asserts a failure immediately. This will mitigate maxinst failing downstream when it's too late to realize that tablespaces and/or bufferzones were not created.

The fix was tested on a live expert labs cluster where they had run into this issue. We retrofitted the fix directly on the cluster and re-executed the role.

https://github.com/ibm-mas/ansible-devops/issues/1155
https://jsw.ibm.com/browse/MASISMIG-48564


ibm/mas_devops/roles/suite_db2_setup_for_manage/tasks/main.yml
```diff
@@ -77,13 +77,20 @@
 - name: Run setupdb to prepare Db2 for use with Maximo TPAE
```
alequint, line 77: @sachbalag be sure to include in PR's.title (we use to ad in any commit messages) the semantic version prefix in squared brackets... in this case it seems a `[minor]` to me as we slightly changed behavior

ibm/mas_devops/roles/suite_db2_setup_for_manage/tasks/main.yml
```diff
@@ -77,13 +77,20 @@
...
+  delay: 30 # seconds
+  until: prepare_cmds_status.rc == 0
 
 - name: "Database setup debug information"
   debug:
     msg:
       - "Result (stdout) ........................ {{ prepare_cmds_status.stdout }}"
       - "result (stderr) ........................ {{ prepare_cmds_status.stderr }}"
 
+- fail: msg="setupdb.sh failed for Manage Maxinst -  tablespace or bufferpools may not have been created)"
```
alequint, line 90: @sachbalag suggestion here is to do the assertion like it is done in other roles, i.g. https://github.com/ibm-mas/ansible-devops/blob/master/ibm/mas_devops/roles/uds/tasks/install/determine-storage-classes.yml#L28 - just to keep the adopted standard/convention used everywhere else

alequint (dismissed):

sachbalag:
The purpose of this PR is to make the db2 setup role more resilient so it fails immediately when there is an issue with the database, rather than failing silently. This fix handles the rc return value from the setupdb.sh script and asserts a failure immediately. This will mitigate maxinst failing downstream when it's too late to realize that tablespaces and/or bufferzones were not created.

The fix was tested on a live expert labs cluster where they had run into this issue. We retrofitted the fix directly on the cluster and re-executed the role.

whitfiea (approved):

PR #1129, opened 2023-12-05
[patch] Install DRO using DRO_STORAGE_CLASS and PVC 

yuvraj-vansure:
DRO should be installed using user supplied Storage class name DRO_STORAGE_CLASS instead of setting a default storage class on the cluster

Code changes:
If DRO_STORAGE_CLASS is supplied, then use the storage class name and create 3 PVC's required for DRO. if DRO_STORAGE_CLASS is not defined or empty then use determine_storeclass role to identify a storage class name and use that to create the PVC's

https://jsw.ibm.com/browse/MASCORE-1097

The issue details were discussed here
https://ibm-watson-iot.slack.com/archives/C040TETRV2T/p1701706299259879

https://ibm-watson-iot.slack.com/archives/C05NBFPNNQ0/p1702484841935989


yuvraj-vansure (commented):
Added Job Block in determine-storage-classes.yml

lokesh-sreedhara (approved):
looks good to me, i checked the Job Block code logic again, it covers all the scenarios we discussed

durera (changes requested):
As discussed in Slack ... we can't be messing with the default storage class as part of the install for a MAS dependency.  We need to get DRO updated to work liek the rest of MAS whereby the storage class can be explicitly set to the one desired rather than have a requirement that the default storage class for the whole cluster is set to whatever storage class DRO needs to use.

lokesh-sreedhara:
Changes complete. DRO role will now use User suppled Storage class name and PVC's to install.

durera (approved):

PR #1130, opened 2023-12-06
[patch] Filter DST X3 Root certificate part from the cluster's ingress chain

andrercm:


terenceq (approved):
These changes look good. With respect to IBM ROKs only the `isrgrootx1` is really required in the truststore (for the ROKs default ingress that is configured)

PR #1135, opened 2023-12-08
[patch] get db2u version from db2u-release configmap instead of secret

sanju7216:
Starting with db2u v`s11.5.8.0-cn3`, the 's' prefix is removed in db2u-license-keys causing installation issues , we were recommended to use `db2u-release` configmap instead of `db2u-license-keys` secret.

This change is in preparation for November curated content from IBM operator catalog. and also backward compatible.

### DIFF
--- a/ibm/mas_devops/roles/db2/README.md
+++ b/ibm/mas_devops/roles/db2/README.md
@@ -75,7 +75,7 @@ Version of the DB2 engine to be used while creating/upgrading the DB2 instances.
 
 - Optional
 - Environment Variable: `DB2_VERSION`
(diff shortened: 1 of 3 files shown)


sanju7216:
this change is verified with new november curated content - https://dashboard.masdev.wiotp.sl.hursley.ibm.com/tests/curatenov

I also checked IVT811-02 which has June curated content and its got the db2u-release configmap, so this shoudl work with older version too.

andrercm:
@sanju7216 Is there any possibility of our automation still support older versions of DB2? If so, will this configmap also be available to be used the same way?

andrercm (approved):
LGTM !

PR #1139, opened 2023-12-12
Change default ipFamilies to IPv4, IPv6 when ipv6_test=true

xinyuwu1:
This PR changes the default ipFamilies to IPv4, IPv6 when ipv6_test=true.


whitfiea:
Closing as inactivity

PR #1153, opened 2023-12-23
[minor] Add support for Red Hat Certificate Manager

andrercm:
**Before merging the PR, complete remaining items:**

- Need to include January catalog as we'll only make Red Hat certmanager switch after January (currently it's set to December in cli)
- Fix `cert_manager` readme to include couple new variables

**PR Overview**

This PR brings Red Hat certificate manager as the new operator to manage and rotate certificates in MAS instances, replacing IBM Certificate Manager (managed by Cloud Pak Foundational Services).

In summary, this should be seamless change for end users while transitioning from IBM Cert Manager to Red Hat corresponding version:

**New MAS installs:**

For new MAS installs, `cert-manager-operator` will be the default namespace to deploy `cert-manager controller` pod and `cert-manager` namespace will be the place to install cert-manager deployment/pods and its corresponding cluster issuer secrets (and therefore also set in Suite CR as `certManagerNamespace: cert-manager`)

image

image

image

image

image

**Existing MAS installs:**

It will automatically detect that IBM cert-manager is installed, configure the pre-reqs for migration to RH cert-manager, then clean up the IBM cert-manager instance. Once, the clean up is done, `cert-manager-operator` will be the default namespace to deploy `cert-manager controller` pod and `cert-manager namespace` will be the place to install cert-manager deployment/pods, however since the cluster issuer secrets will already be present under `ibm-common-services` namespace, we'll set both CertManager CR and Suite CR to still point to `ibm-common-services` namespace (in Suite CR,`certManagerNamespace: cert-manager`) so that **we don't need** to migrate the existing certificates and corresponding secrets to a new `cert-manager` namespace.


durera (changes requested):
As discussed in Slack, just need to flip the default and we can get this merged in.

durera (approved):

PR #1161, opened 2024-01-08
[patch] change to use SingleStack and IPv6 only when ipv6_test is true

xinyuwu1:
The default ipFamilyPolicy and ipFamilies are now changed to SingleStack and IPv6 to better test IPv6-only configurations.

### DIFF
--- a/ibm/mas_devops/roles/suite_install/README.md
+++ b/ibm/mas_devops/roles/suite_install/README.md
@@ -87,7 +87,7 @@ For examples refer to the [BestEfforts reference configuration in the MAS CLI](h
 - Default: None
 
 ### enable_IPv6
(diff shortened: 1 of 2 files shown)


whitfiea (commented):
@xinyuwu1 what is the reason for changing the default? I am not sure I understand why we would be making changes for tests into the master branch?

xinyuwu1:
> @xinyuwu1 what is the reason for changing the default? I am not sure I understand why we would be making changes for tests into the master branch?

xinyuwu1:
@whitfiea Originally I had the ipFamily properties for services set to PreferDualStack and ["IPv6", "IPv4"] .  As discussed in Peter's architecture call yesterday, with these settings the fvt test doesn't uncover any problem specific with IPv6 as it can always use IPv4.  So we are changing the default setting to SingleDualStack and ["IPv6"] when the ipv6_test is true.

These are not changes for tests, but the default settings for the ipFamily properties in services.

xinyuwu1:
@whitfiea The change is just to provide a different ipfamily setting for the Suite CR when test_ipv6 is true.  Please review again.

whitfiea (approved):
As this can only be used for internal clusters for now then I approve. But a breaking change like this would normally need a major tag to indicate that the default has changed.

PR #1165, opened 2024-01-11
[patch] Fixes for UDS migration to DRO scenario

lokesh-sreedhara:
MAS migration to DRO
 - changes to docs to update new env variables needed for DRO
 - DRO_MIGRATION flag added to handle the migrations from UDS to DRO via `mas update `on cli. the cli will pass down the value for this flag
 - UDS will be uninstalled when DRO_MIGRATION is true
 - dro will be installed when DRO_MIGRATION is true
 - gencfg tasks will look at all available bascfgs, recreate new onces with DRO if DRO_MIGRATION flag is true 

related CLI changes are here: https://github.com/ibm-mas/cli/pull/726


whitfiea (dismissed):
@lokesh-sreedhara This PR needs a description

ibm/mas_devops/roles/dro/tasks/gencfg/migrate.yml
```diff
@@ -0,0 +1,68 @@
+---
+- name: "gencfg-migrate : debug UDS BASCfgs"
+  debug:
+    msg:
+      - "name:        {{item.metadata.name}}"
```
andrercm, line 5: Can we standardize the debug following other samples as in https://github.com/ibm-mas/ansible-devops/blob/master/ibm/mas_devops/roles/cp4d/tasks/install-cp4d.yml#L46

andrercm (changes requested):
Just minor feedback on the debug statement.
Plus, following David's chat we need to ensure uds continues as default option until Feb patch release is out.

ibm/mas_devops/roles/dro/tasks/gencfg/migrate.yml
lokeshs09: @andrercm done. i have updated the debug msg

lokeshs09:
> Plus, following David's chat we need to ensure uds continues as default option until Feb patch release is out.

how do we ensure this? should i remove the entry for dro role from the oneclick_core?

ibm/mas_devops/roles/dro/tasks/gencfg/migrate.yml
lokesh-sreedhara: @andrercm done. i have updated the debug msg

docs/playbooks/oneclick-core.md
durera, line 1: It's too early to update the doc to talk about DRO ... once merged this doc will be live, it needs to stay the same - UDS.  The change to make DRO the default needs to be introduced in the branch that will be created for the Feb catalog update when DRo is planned to go live.

ibm/mas_devops/playbooks/oneclick_core.yml
durera, line 1: The playbook before the changes here is fine, UDS can't just disappear like it never existed even after we add DRO, this still needs to work is a user wants to install an older catalog version (which would result in UDS instead of DRO)

ibm/mas_devops/roles/dro/README.md
```diff
@@ -59,6 +52,34 @@ Local directory to save the generated BasCfg resource definition.  This can be u
 - Environment Variable: `MAS_CONFIG_DIR`
 - Default Value: None
 
+### dro_existing
```
durera, line 55: What's the purpose of this?  In this scenario I would expect teh action to be that the user sets `uds_action=none` so that DRO is not installed, and pass in the required BASCfg with the addtional configurations mechinism supported in suite_config, so that this is consistent with how you would configure an external mongo, db2, or kafka instance.  We don't need to create another code path that needs to be maintained/tested for achieving the same thing.

durera (changes requested):

ibm/mas_devops/playbooks/oneclick_core.yml
lokesh-sreedhara: i have reverted the playbook and playbook docs to how it was before.

docs/playbooks/oneclick-core.md
lokesh-sreedhara: i have reverted the playbook and playbook docs to how it was before.

ibm/mas_devops/roles/dro/README.md
durera: CLI does not support BYO UDS, DRO, SLS, or Mongo today.  These installations are only supporting by using the ansible roles/building your own playbook.

durera: If we want to add support for them refer to the pattern established for the JDBCCfg management, whereby the user is promoted to generate the configuration and it's added to the additional configs metadata.

lokesh-sreedhara: @durera yea the BYO DRO is support on ansible roles. CLI does not have it. do u expect dro_existing to be removed from the ansible role as well?

lokesh-sreedhara: looks like @yuvraj-vansure already removed it.

durera (approved):

PR #1167, opened 2024-01-11
[patch] add licensing sync cronjob sync frequency in slscfg

conradkao:
Make the licensing sync cronjob sync frequency configable in slscfg.
The default value is  30 min `'*/30 * * * *'`


terenceq (changes requested):
@conradkao does the new env var `MAS_LICENSE_SYNC_FREQUENCY` need to be documented here:


https://github.com/ibm-mas/ansible-devops/tree/master/ibm/mas_devops/roles/sls

It appears to be a cron configuration and I think it would be good to explain what it is and what the default value is/does

terenceq (approved):

PR #1170, opened 2024-01-16
[patch] ObjectStorage config should have type

whitfiea:
The ObjectStorageConfig being created in the `cos` role should have the `type` set in the spec. This was failing the FVT tests when an existing config was attempted to be reapplied, as the check for the `type` failed. Tested again in https://dashboard.masdev.wiotp.sl.hursley.ibm.com/tests/pfvtajw1304 and this fixes the internalapi tests.


andrercm (approved):
LGTM

durera (approved):

Issue #978, opened 2023-08-22
Grafana is installed with anonymous access set as default 


andrercm:
This is the location - https://github.com/ibm-mas/ansible-devops/blob/master/ibm/mas_devops/roles/cluster_monitoring/templates/grafana.yml.j2

Action:
The default Grafana option for authentication option of Grafana should not allow anonymous access.
Reference: https://grafana.com/docs/grafana/latest/setup-grafana/configure-security/configure-authentication/grafana/

Additional discussion in Slack - https://ibm-watson-iot.slack.com/archives/C040TETRV2T/p1692116049456159

stonepd:
fixed in https://github.com/ibm-mas/ansible-devops/pull/1188

durera:
Closing issue, this is now resolved

PR #1188, opened 2024-01-25
[minor] Support Grafana Operator v5

stonepd:
## Description
Provide capability to install Grafana Operator V5, compatible datasources and dashboards.
Provide capability to uninstall Grafana Operator V4, and all related resources.
Disable anonymous access to grafana

## Related Issues
epic https://jsw.ibm.com/browse/MASCORE-1270
https://jsw.ibm.com/browse/MASCORE-1283
https://jsw.ibm.com/browse/MASCORE-1381

https://github.com/ibm-mas/ansible-devops/issues/978

## How Has This Been Tested?
This has been tested in `pfvtpds` personal FVT cluster.


istrate (dismissed):
looks good!

ibm/mas_devops/roles/cluster_monitoring/tasks/update_grafana/main.yml
```diff
@@ -0,0 +1,49 @@
+---
+# 1. Load default storage classes (if not provided by the user)
+# -----------------------------------------------------------------------------
+- name: "update_grafanav4 : Determine storage classes"
```
whitfiea, line 4: `update_grafanav4` should be `update_grafana` as that is the task name. I think this similar issue is seen in a few other task files where the name doesn't match the task it is now in.

ibm/mas_devops/roles/cluster_monitoring/README.md
```diff
@@ -13,7 +13,7 @@ The credentials for the grafana admin user are stored in `grafana-admin-credenti
 Role Variables
 -------------------------------------------------------------------------------
 ### cluster_monitoring_action
-Inform the role whether to perform an install or an uninstall of cluster monitoring.
+Inform the role whether to perform an install or an uninstall of cluster monitoring. Can also be set to update_grafana to update the Grafana Operator from V4 to V5.
```
whitfiea, line 16: Does the update keep the login data that might be setup? I don't think it would as the PVC would be new and in a different namespace. So if uses have setup users in v4 then updating to v5 means those users are lost?

whitfiea (changes requested):

ibm/mas_devops/roles/cluster_monitoring/tasks/update_grafana/main.yml
stonepd: Thanks, now changed

ibm/mas_devops/roles/cluster_monitoring/README.md
durera: I think we can document it as a side-effect of the upgrade.

> When Grafana v5 is installed it will not inherit the user database from the old v4 installation, the admin password will be new, and user accounts set up in the v4 instance will need to be recreated in the v5 instance.

stonepd: OK, does this make it a major release?

stonepd: @durera , @whitfiea - I've updated the readme

durera: Nah, we're still fine as a minor

durera (approved):

PR #1212, opened 2024-02-19
[minor] Added failure handling for Maxinst setup script in suite_db2_setup_for_manage role

sachbalag:
The purpose of this PR is to make the db2 setup role more resilient so it fails immediately when there is an issue with the database, rather than failing silently. This fix handles the rc return value from the setupdb.sh script and asserts a failure immediately. This will mitigate maxinst failing downstream when it's too late to realize that tablespaces and/or bufferzones were not created.

https://jsw.ibm.com/browse/MASISMIG-49684


ibm/mas_devops/roles/suite_db2_setup_for_manage/tasks/main.yml
```diff
@@ -77,13 +77,30 @@
...
+    msg:
+      - "SetupDB return code (rc) ........................ {{ prepare_cmds_status.rc }}"
 
 - name: "Database setup debug information"
   debug:
     msg:
       - "Result (stdout) ........................ {{ prepare_cmds_status.stdout }}"
       - "result (stderr) ........................ {{ prepare_cmds_status.stderr }}"
 
+- name: Assert that manage tablespace or bufferpools are not created
```
whitfiea, line 96: Shouldn't this say "Assert that manage tablespace or bufferpools are created" rather than "Assert that manage tablespace or bufferpools are **not** created"

whitfiea (changes requested):

ibm/mas_devops/roles/suite_db2_setup_for_manage/tasks/main.yml
sachbalag: Yes, you're right!

whitfiea (approved):

PR #1221, opened 2024-02-26
[patch] change dro_migration flag to lower case 

yuvraj-vansure:
https://jsw.ibm.com/browse/MASCORE-1825

- added lower case check for dro_migration variable.
- Ensure UDS role does not install during mas update


lokesh-sreedhara (changes requested):
- `dro_migration == "" or dro_migration | lower == "false"`

We need the above condition in UDS role, to ensure it does not run install again during mas update. 

This will ensure uds install is skipped and control moves on to dro role.

lokesh-sreedhara (dismissed):

sanju7216 (approved):

PR #1233, opened 2024-03-07
[major] Remove deprecated capabilities

durera:
Removes:
- [x] Airgap v1 support (digest configmaps)
- [x] Support for end of life MAS releases (MAS 8.6 to 8.9)
- [x] Support for catalogs prior to MAS 8.10 GA In April 2023
- [x] `convert_to_olm`, `suite_mustgather`, `suite_mustgather_download`, `suite_install_digest_cm` roles
- [x] `convert_to_olm` playbook
- [x] Remove `cluster_monitoring` role, function replaced by `grafana` and `opentelemetry` in their own roles, and move the OCP monitoring configuration code into existing `ocp_config` role


durera:
> There is still references to the convert_to_olm role in the MAS docuemntation page https://www.ibm.com/docs/en/mas-cd/continuous-delivery?topic=versions-converting-manual-deployment-channel-subscription

We should get that cleaned up, ideally removed, or failing that referencing a fixed version of the collection that still supports the conversaion.  The supported releases of MAS are all OLM only at this point.

durera:
> What is the alternative for the suite_mustgather role, as that might be used must by any users.

https://ibm-mas.github.io/cli/commands/must-gather/

We replaced the old must-gather with this one a while back now, I've not seen any must-gathers in the old format for a while, and if we did we would respond asking them to run the new must-gather rather than the old one that is used in these roles.

PR #1234, opened 2024-03-08
[minor] Support Cloud Pak for Data 4.8.x - Install & Upgrade

andrercm:


ibm/mas_devops/roles/dro/README.md
durera, line 1: Think this is a search/replace gone wrong ... see other files with similar as well

andrercm: fixed

durera (approved):

PR #1237, opened 2024-03-13
[patch] add health dbc loader

MollyGu-CN:


whitfiea (commented):
@MollyGu-CN The PR needs a description of what this is adding and also added to the documentation page. You also need to show the testing done. It should also be a minor rather than patch as this is adding function

MollyGu-CN:
> @MollyGu-CN The PR needs a description of what this is adding and also added to the documentation page. You also need to show the testing done. It should also be a minor rather than patch as this is adding function

Per @andrercm suggestion, I will try modify this a little bit, make it can work both manage with health and health standalone env

PR #1239, opened 2024-03-14
[minor] Extend 'suite_manage_load_dbc_scripts' role to support health standalone

MollyGu-CN:
Add a variable MAS_APP_ID to support health standalone case.

Tested in one of the health standalone env

image

image


ibm/mas_devops/roles/suite_manage_load_dbc_scripts/tasks/main.yaml
```diff
@@ -49,6 +49,7 @@
       debug:
         msg:
           - "Instance ID ............... {{ mas_instance_id }}"
+          - "App ID ............... {{ mas_app_id }}"
```
andrercm, line 52: Would you mind fixing the standard dots in the debug just to keep it consistent with other lines?

ibm/mas_devops/roles/suite_manage_load_dbc_scripts/defaults/main.yml
```diff
@@ -0,0 +1,2 @@
+
```
andrercm, line 1: would you be able to add an extra interna var, something as:
```
mas_app_id_supported:
  - 'manage'
  - 'health'
```

and then add an assertion right at the beginning of the `tasks/main.yml` that validates that the passed `mas_app_id` matches one of the `mas_app_id_supported`?

Here's a sample of how you'd do it:
https://github.com/ibm-mas/ansible-devops/blob/master/ibm/mas_devops/roles/cp4d/tasks/main.yml#L13

ibm/mas_devops/roles/suite_manage_load_dbc_scripts/vars/main.yaml
```diff
@@ -1,7 +1,9 @@
 ---
```
andrercm, line 1: It doesn't seem good practice to have default vars in `vars/main.yaml` file... not sure why this was done like this in first place... instead we should get rid of this folder/file and move all the default vars to the new `defaults/main.yml` you created.

andrercm (changes requested):
Can you work on the couple of changes ?

ibm/mas_devops/roles/suite_manage_load_dbc_scripts/tasks/main.yaml
MollyGu-CN: Added

ibm/mas_devops/roles/suite_manage_load_dbc_scripts/defaults/main.yml
MollyGu-CN: Added and tested
image

ibm/mas_devops/roles/suite_manage_load_dbc_scripts/vars/main.yaml
MollyGu-CN: Moved

MollyGu-CN: vars/main.yaml is deleted, but the folder is still there? I have removed entire folder in local, but it doesn't appear in the commit...

andrercm (approved):
Thanks for making the changes

PR #1247, opened 2024-03-25
[minor] Extend support for Manage additional settings

andrercm:
The following changes are requests for enhancements discussed and raised with a business partner that is actively using our MAS cli.

This PR adds support for some additional Manage settings:

Base and secondary languages
- Allow Manage dedicated Db2 instance to be configured with `db2oltp` type instead of default `db2wh`. The `db2wh` will continue to be the default option, although allowing `db2oltp` for Manage might be useful for customers using Manage & Cognos integration in the same database since Cognos does not support `db2wh` type.
- Server timezone: Allow a timezone to be set at ManageWorkspace CR level with `{{ mas_app_settings_server_timezone }}` property. In addition, allow defining same value in `spec.advOpts.timezone: {{ db2_timezone }}` property for `DB2Ucluster` in `db2` role. This way, both Manage server and the corresponding DB2 instance will be running within the same timezone.

Once this is merged, we need to merge MAS CLI PR: https://github.com/ibm-mas/cli/pull/854


ibm/mas_devops/roles/db2/defaults/main.yml
```diff
@@ -12,6 +11,9 @@ db2_channel: "{{ lookup('env', 'DB2_CHANNEL') }}"
 
 db2_instance_name: "{{ lookup('env', 'DB2_INSTANCE_NAME') }}" # e.g. db2u-iot or db2u-manage
 db2_dbname: "{{ lookup('env', 'DB2_DBNAME') | default('BLUDB', true) }}"
+db2_type: "{{ lookup('env', 'DB2_TYPE') | default('db2wh', true) }}" # e.g. db2wh or db2oltp
+db2_timezone: "{{ lookup('env', 'DB2_TIMEZONE') }}"
```
whitfiea, line 15: Are we missing a readme entry for db2_timezone?

whitfiea (changes requested):

ibm/mas_devops/roles/db2/defaults/main.yml
andrercm: thanks! just fixed !

whitfiea (approved):

PR #1248, opened 2024-03-25
[patch] Added pipefail to ensure piped command's RC is not masked

sachbalag:
This fix adds "pipefail" to the "oc exec" command pipeline and will ensure that if either one of the commands in the pipeline, "oc exec" or "tee" fails, the entire pipeline will fail, and a non-zero exit code is returned. This will fix error handling issues downstream and accurately assert a failure if "setupdb.sh" fails.

https://jsw.ibm.com/browse/MASISMIG-52381


whitfiea (changes requested):
@sachbalag was this change tested?

sachbalag:
@whitfiea Yes this was tested on a QB cluster. Here is the link to the log file with the results ...

https://jsw.ibm.com/secure/attachment/8060026/MASISMIG-52381%20-%20log.txt

whitfiea (approved):

PR #1253, opened 2024-03-28
[patch] apply same cp4d zen fix for cp4d_service role

sanju7216:
Missed to add the zen fix in cp4d_service role.
So reusing the fix here from cp4d role.


andrercm (approved):
approved

PR #1262, opened 2024-04-05
[major] Delete all CertManager CRDs with cert-manager uninstall

andrercm:
This PR addresses problems found in https://github.com/ibm-mas/cli/issues/868

It is all about including Certificate Manager CRD to be deleted in `cert_manager role` during `uninstall` action. Without that, during common services reinstall, ibm common services operator will lookup Certificate CRD and will find it and think certmanager is installed and call cert manager webhook pod (which is not present because it was deleted during certmanager uninstall) and then fail to proceed with ibm-common-services reinstall...

Tests done:

image

CertManager CRDs before uninstall:

image

All Certmanager CRDs are  gone:

image

Then, re-running `mas install` command after a `mas uninstall` allows reinstalling ibm common services and cert-manager just fine:

image


whitfiea (approved):

whitfiea:
Changed this to major as it changes the behaviour and could cause issues for users that are expecting the old behaviour of the CRDs to be left.

PR #1269, opened 2024-04-12
[patch] Update wd.yml.j2 to enable disableStsConflictCheck

liyongc:
In some Env, Watson Discovery hdp worker repeatedly restarted during the document indexing intermettently, and based on the comments from WD support team and seems the `disableStsConflictCheck` should be set true to work-around the issue.


andrercm:
@liyongc is there a test evidence showing that WD install completes successfully with this setting enabled? Cheers.

whitfiea:
@liyongc any update?

liyongc:
No plan to maintain WD in ansible-devops since WD is removed from MAS 9.0.

PR #1270, opened 2024-04-12
[patch] Add norootsquash for ibm-vpc-file

whitfiea:
The norootsquash daemonset needs to be present for an NFS based storage. A condition was added to only install this for the classic IBM Cloud ibmc-file storage but this is also needed for the IBM Cloud VPC file storage as well

This was tested in an IBM Cloud VPC environment


andrercm (approved):
LGTM

PR #1275, opened 2024-04-18
[minor] [MASCORE-1495] Support CP4D v4.8 in Airgap cluster

andrercm:
In this PR:
- Prepare the mirroring roles to include CPD 4.8 images
- Add Cognos Analytics as part of the mirroring of dependencies


whitfiea (changes requested):
@andrercm This seems to have removed the cp4d 4.8.1 mirror_extra_perpare vars but the catalog_source replacement is just 4.8.0. What happens if the CPD_PRODUCT_VERSION is 4.8.1?

andrercm:
@whitfiea at this point we are just going to support CPD 4.8.0, once it's in, then we will later on decide which CPD patch to support, that's why i removed 4.8.1 to avoid confusion on which version to use.

sanjayprab (approved):

whitfiea (approved):

PR #1289, opened 2024-05-03
CPD 4.8.0 casebundle definition

andrercm:
**Keep in draft until CPD 4.8 is formally claimed support**

Details in this PR corresponds to Cloud Pak For Data 4.8.0 bundle image tags to be mirrored.
Since we are uncertain still regarding which monthly patch this will be claimed support, we need to add the details contained here to whatever catalog / case bundle file that starts enabling CPD 4.8.0 to be installed by default


sanjayprab:
This has been added to june catalog. So closing it.

PR #1291, opened 2024-05-03
[patch] fix iamIntegration default to false and adjust default backin…

andrercm:
…gStorage resources

### DIFF
--- a/ibm/mas_devops/roles/cp4d/defaults/main.yml
+++ b/ibm/mas_devops/roles/cp4d/defaults/main.yml
@@ -3,6 +3,7 @@
 cpd_primary_storage_class: "{{ lookup('env', 'CPD_PRIMARY_STORAGE_CLASS') }}"
 cpd_metadata_storage_class: "{{ lookup('env', 'CPD_METADATA_STORAGE_CLASS') }}"
 cpd_scale_config: "{{ lookup('env', 'CPD_SCALE_CONFIG') | default('medium', true) }}"
+cpd_iam_integration: "{{ false | bool }}"
 
 # ------ CPD entitlement ----------------------------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


ibm/mas_devops/roles/cp4d_service/tasks/prereqs/odf/storage.yml
```diff
@@ -75,7 +74,29 @@
     - backingstore_info.resources | length > 0
     - backingstore_info.resources | json_query('[*].status.conditions[?type==`Available`][].status') | select ('match','True') | list | length == 1
 
-# 4. Provide debug information
+# This is needed because sometimes the noobaa-default-backing-store-noobaa-pod does not get started correctly
+# so restarting all corresponding backing store operators/pods seems to do the trick
+- name: "Restart all pods from openshift-storage namespace"
+  shell: oc delete pods --all -n openshift-storage
```
whitfiea, line 80: This seems quite serve to do, and it seems it will do this on each run of the role even if everything is ok. I assume that the deletion of all these pods will cause a storage outage? or would it only effect new storage provisioning?

whitfiea (changes requested):

ibm/mas_devops/roles/cp4d_service/tasks/prereqs/odf/storage.yml
andrercm: So this is only going to happen during the first provisioning of the BackingStore/MCG... but you made a very good point... i made a fix where instead of deleting all pods from the namespace, I'm now just deleting the `noobaa-operator` pod, which in fact is the one responsible for putting up the missing pod that is required.

whitfiea (approved):

PR #1293, opened 2024-05-06
[patch] Fix lookup kafka subscription

andrercm:
This PR just fixes a minor bug in the strimzi subscription where we were not waiting for `status` property to be present before moving forward.


cuddlyporcupine (commented):
Looks good to me

cuddlyporcupine (approved):

whitfiea (approved):

PR #1298, opened 2024-05-07
[minor] Make DRO default in oneclick_core playbook

lokesh-sreedhara:
- switch bas_provider to install DRO by default from onclick_core playbook 
- Adjust the DRO contact env variables to look up UDS onces by default if they are not provided by the user to ensure the old flow does not break and users dont have to modify any existing automation.

### DIFF
--- a/ibm/mas_devops/playbooks/oneclick_core.yml
+++ b/ibm/mas_devops/playbooks/oneclick_core.yml
@@ -34,24 +34,12 @@
         fail_msg: "One or more required environment variables are not defined"
 
     - name: Check for required environment variables for UDS
-      when: (bas_provider =='UDS')
       assert:
         that:
           - lookup('env', 'UDS_CONTACT_EMAIL') != ""
           - lookup('env', 'UDS_CONTACT_FIRSTNAME') != ""
           - lookup('env', 'UDS_CONTACT_LASTNAME') != ""
         fail_msg: "One or more required environment variables are not defined"
-
-
-    - name: Check for required environment variables for DRO
-      when: (bas_provider =='DRO')
-      assert:
-        that:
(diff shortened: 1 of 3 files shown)


benbakowski (approved):
As discussed on slack!

ibm/mas_devops/playbooks/oneclick_core.yml
```diff
@@ -34,25 +34,13 @@
         fail_msg: "One or more required environment variables are not defined"
 
     - name: Check for required environment variables for UDS
-      when: (bas_provider =='UDS')
       assert:
```
whitfiea, line 37: This changes means the user has to set the UDS_X variables even if this is a new user that wants DRO and doesn't know what UDS is. I think we should put back this when condition

ibm/mas_devops/playbooks/oneclick_core.yml
```diff
@@ -34,25 +34,13 @@
...
 
-
-    - name: Check for required environment variables for DRO
-      when: (bas_provider =='DRO')
-      assert:
-        that:
-          - lookup('env', 'DRO_CONTACT_EMAIL') != ""
-          - lookup('env', 'DRO_CONTACT_FIRSTNAME') != ""
-          - lookup('env', 'DRO_CONTACT_LASTNAME') != ""
-        fail_msg: "One or more required environment variables are not defined"
```
whitfiea, line 53: Can we put this back but update the check so it checks for both i.e.

```
lookup('env', 'DRO_CONTACT_EMAIL') != "" or lookup('env', 'UDS_CONTACT_EMAIL') != ""
```
note: I haven't tested this line so it might need tweaking to get it working

whitfiea (changes requested):
As we are changing the default then we should probably make this a minor.

durera (approved):

PR #1314, opened 2024-05-20
[patch] include grafana namespaces in image source policy

stonepd:
Grafana V5 operator fails to install in the airgap mode because the operator bundle image location is not correctly mapped in the Image Source Content Policy.

Fixing https://jsw.ibm.com/browse/MASCORE-2737

Tested in airgap FVT (see https://dashboard.masdev.wiotp.sl.hursley.ibm.com/tests/fvtairgap/testsuite/ibm-mas-devops/setup-cluster-monitoring, `roles/cluster_monitoring/tasks/install/grafana-v5` subsection). 

I enabled Grafana Install, Run 1210 was fixed with manual changes and Run 1212 was with fixed branch image.
image


ibm/mas_devops/roles/ocp_contentsourcepolicy/templates/redhat-catalogs.yml.j2
```diff
@@ -55,6 +55,16 @@ spec:
   - mirrors:
     - {{ registry_private_url }}/grafana-operator
```
sanjayprab, line 56: Do we still need it? isn't this v4 which doesn't work on airgap?

sanjayprab (approved):

PR #1339, opened 2024-06-13
[minor] Support NFS role and enabled it by default on Fyre clusters

jeancrg:
- This PR adds the following:

- changed the default values for the cluster sizing to minimum enough when using NFS
- added FYRE_WORKER_ADDITIONAL_DISK variable to set the additional disks in the worker nodes only when needed, since NFS does not need any, and if letting the default value, the cluster OCP+ always is created with 400Gi.

- created the NFS role to be used only for Fyre IT

- updated provision fyre playbook to support NFS, making it default when installing through cli

- Tests: 

  Tested various NFS installations using Fyre Quick Burn, with a different combination of variables, for example, installing by passing the FYRE_APIKEY and FYRE_USERNAME or installing by passing the FYRE_INF_NODE_PRIVATE_IP.


andrercm (approved):
LGTM, thanks JC for putting this together!

durera (approved):

PR #1342, opened 2024-06-18
[minor] DRO should default to ibm-software-central namespace 

lokesh-sreedhara:
"DRO operator v2.16.0 is published and defaults namespace to` ibm-software-central`. the role bindings are created for that namespace, as well as the legacy default namespaces so, should not impact upgrade, and for new deployments, you can use` ibm-software-central` namespace if you wish."

This PR is to change the DRO default namespace from `redhat-marketplace` to ` ibm-software-central` 

Related CLI PR: https://github.com/ibm-mas/cli/pull/1040

### DIFF
--- a/ibm/mas_devops/roles/dro/README.md
+++ b/ibm/mas_devops/roles/dro/README.md
@@ -5,7 +5,10 @@ DRO will be supported on the following MAS versions
 - MAS 8.11.2 +
 - MAS 9.0 +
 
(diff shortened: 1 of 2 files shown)


yuvraj-vansure (approved):

whitfiea (changes requested):
@lokesh-sreedhara This would have to be a major change I believe as you are changing the default. If someone ran this role before this change, and then runs it after this change on the same environment, then you would have DRO installed twice? We would need some type of migration path so we don't duplicate deployments

lokesh-sreedhara:
> @lokesh-sreedhara This would have to be a major change I believe as you are changing the default. If someone ran this role before this change, and then runs it after this change on the same environment, then you would have DRO installed twice? We would need some type of migration path so we don't duplicate deployments

@whitfiea hmm. yea if it would install it twice if users install old version and then the newer version. thinking how to handle this.

whitfiea:
https://github.com/ibm-mas/cli/pull/1040 is closed as won't do, so closing this as well

PR #1360, opened 2024-07-01
[patch] Add dro to uninstall

sanjayprab:
replace uds uninstall with dro uninstall in uninstall playbook.


whitfiea (changes requested):
Can we not have both? Incase someone is uninstalling a version that has UDS?

whitfiea (approved):

PR #1361, opened 2024-07-02
[minor] Support for resource management via gitops

whitfiea:
The following changes are required to support [gitops](https://github.com/ibm-mas/gitops) approach:

- Adds username/password to wipe_mongo.py so that it can be called for a docdb instance.
- Creates a specific action on the mongodb role called `destory-data` that calls the wipe_mongo library
- For CIS dns entries it diables the CIS Proxy for IoT messaging endpoints due to CIS not supporting MQTT
- Fixes some issues with the rosa provisioning role
- `suite-certs` role takes a GITOPS flag that if true will do all the CIS parts but not install any resources on the OCP cluster.

The above changes run in fvtsaas today. Also tested in a non-gitops env:
image- 2024-07-02 at 17 04 46


durera (commented):
Review comments discussed in Slack

whitfiea:
Added changes to make a destory-data action as part of the mongodb aws provider role, rather than have a specific role for it.

durera (approved):

PR #1377, opened 2024-07-08
[patch] Remove common_services role from oneclick_core 

sanjayprab:
Common_services role is not required for core any more. it serves no purpose currently. 
Common_services is only installed as part of cp4d. CP4D role will install common-services v4.3.0.

This will resolve airgap issue.


terenceq (approved):

sanserifmsk:
Common services creates CRD OperandRequest, that used in next steps for installing cert manager in https://github.com/ibm-mas/ansible-devops/blob/master/ibm/mas_devops/roles/cert_manager/templates/ibm/ibm-cert-manager.yml

PR #1391, opened 2024-07-22
[minor] Add repository prefix support for image mirroring

JonahLuckett:
### Overview

For https://github.com/ibm-mas/cli/issues/1049

With this change the user can declare a Registry Prefix that will publish the mirror image created to `HOST:PORT/PREFIX/ `as opposed to `HOST:PORT/`

There is an accompanying [CLI change](https://github.com/ibm-mas/cli/pull/1130) also but it should be worth note that neither will break the other if not present. Equally if no prefix is given then nothing breaks either.

The prefix is added to a with a `-x` or a `--prefix` flag.

### Testing

Testing in masdeps2 resulted in succesful pod/mirror creation:

```
(B Target Registry(B
(B Registry Host ....................... airgap-registry-lb.airgap-registry.svc
(B Registry Port ....................... 5000
(B Registry Prefix ..................... 240625

(B IBM Operator Catalog(B
(B Catalog Version ..................... v9-240625-amd64
(B MAS Update Channel .................. 9.0.x
```

Screenshot 2024-08-07 at 11 21 37

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,0 +1,56 @@
+name: 🐛 Bug Report
+description: File a bug report
+labels: [Bug Report]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Thanks for taking the time to fill out this bug report!
+
+  - type: input
+    id: version
+    attributes:
+      label: Collection version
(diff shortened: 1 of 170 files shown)


JonahLuckett:
Doing final registry checks before merging

durera (approved):

Issue #1378, opened 2024-07-10
OpenshiftSDN is deprecated


faangbait:
Per [RedHat](https://docs.redhat.com/en/documentation/openshift_container_platform/4.15/html/release_notes/ocp-4-15-release-notes#ocp-4-15-bmer): 

>11.1.1.1. SDN
>
>The SDN controller is not supported with single-node OpenShift.
>The SDN controller does not support IPv6.
>The SDN controller is not supported for OpenShift Container Platform 4.15 and later releases.

The ocp_provision role provides no way to prefer OVNKubernetes in IPI installs. https://github.com/ibm-mas/ansible-devops/blob/6b2efe919b554a09fb74aa1dc1bf9c9ae3960f35/ibm/mas_devops/roles/ocp_provision/templates/ipi/install-config.yaml.j2#L27

JonahLuckett:
Is being solved by https://github.com/ibm-mas/ansible-devops/pull/1403 - will close this issue once that is merged in

PR #1403, opened 2024-08-02
[PATCH] remove SDN networkpolicy

IanBoden:
openshiftSDN networktypes have been removed in OCP 4.15 so we should always use OVNkubernetes


JonahLuckett:
**_Worth note:_** Solves https://github.com/ibm-mas/ansible-devops/issues/1378

durera (approved):

PR #1404, opened 2024-08-02
[patch] Use correct name of k8s_info library

faangbait:
Similar verification steps use `kubernetes.core.k8s_info` instead of `community.kubernetes.k8s_info`


durera (approved):

durera:
Thanks, will get this fix included in the next release

PR #1405, opened 2024-08-07
[patch] MASCORE-3279- Manage attachment causes failure even though it succeed as part of manage-cfg third step

chiragcsatani:
### Description:
- Added retries and delay for task which uses until
- Added new task at bottom to include this task if condition is met

### Related Links:
https://jsw.ibm.com/browse/MAXMF-445

### Testing:

Tested in a personalfvt
Logs: [app cfg manage.log](https://github.com/user-attachments/files/16787035/app.cfg.manage.1.log)
Dashboard: https://dashboard.masdev.wiotp.sl.hursley.ibm.com/tests/pfvtalefq

Screenshot 2024-08-28 at 23 06 09


alequint (changes requested):
@chiragcsatani the pipeline with this PR change has failed here. Take a loot at the run 3195 of the personal environment it was tested: https://dashboard.masdev.wiotp.sl.hursley.ibm.com/tests/pfvtalefq/testsuite/ibm-mas-devops/app-manage-cfg

```
      "msg": "The conditional check 'ready_bundle_count != bundle_count and bundle_retry_count < 60' failed. The error was: Unexpected templating type error occurred on ({% if ready_bundle_count != bundle_count and bundle_retry_count < 60 %} True {% else %} False {% endif %}): '<' not supported between instances of 'str' and 'int'. '<' not supported between instances of 'str' and 'int'\n\nThe error appears to be in '/opt/app-root/lib64/python3.9/site-packages/ansible_collections/ibm/mas_devops/roles/suite_manage_attachments_config/tasks/common/wait-bundle-pods.yml': line 51, column 7, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n    - name: \"wait : Wait for 30 seconds before checking again\"\n      ^ here\nThis one looks easy to fix. It seems that there is a value started\nwith a quote, and the YAML parser is expecting to see the line ended\nwith the same kind of quote. For instance:\n\n    when: \"ok\" in result.stdout\n\nCould be written as:\n\n   when: '\"ok\" in result.stdout'\n\nOr equivalently:\n\n   when: \"'ok' in result.stdout\"\n"
```

alequint (approved):

PR #1412, opened 2024-08-13
[minor] Add "Special Characters" support to MAS install

RexMCM:
## Description
Adding support for the new special_character parameter

## Related Issues

- [MASCORE-3417](https://jsw.ibm.com/browse/MASCORE-3417)

## How Has This Been Tested?
I conducted a full-person functional verification test (FVT) on the entire development branch, including the CLI.
image
 
Also, used CLI to install a new instance using these development branches to make sure the manual process will work correctly:
image
image


ibm/mas_devops/roles/suite_install/defaults/main.yml
```diff
@@ -9,6 +9,9 @@ mas_channel: "{{ lookup('env', 'MAS_CHANNEL') }}"
 mas_domain: "{{ lookup('env', 'MAS_DOMAIN') }}"
 mas_instance_id: "{{ lookup('env', 'MAS_INSTANCE_ID') }}"
 
+# Special characters
```
durera, line 12: This needs to cater for older versions of MAS where this settings does not exist.

e.g. should default to undefined/""

```
mas_special_characters : "{{ lookup('env', 'MAS_SPECIAL_CHARACTERS') }}"
```

User must set it to true|false, if they don't it's "", which means we do nothing.

ibm/mas_devops/roles/suite_install/templates/core_v1_suite.yml.j2
```diff
@@ -83,6 +83,8 @@ spec:
 {% if mas_channel != '8.7.x' and mas_channel != '8.8.x' %}
     manualCertMgmt: {{  mas_manual_cert_mgmt  }}
 {% endif %}
+    userDataValidation:
```
durera, line 86: Wrap this in a `{% if mas_special_characters != '' %}` conditional so that we only add it when the user chooses to set this config option

durera (changes requested):

ibm/mas_devops/roles/suite_install/defaults/main.yml
RexMCM: "This should be sufficient, right? If we're validating the condition in the [ibm/mas_devops/roles/suite_install/templates/core_v1_suite.yml.j2](https://github.com/ibm-mas/ansible-devops/pull/1412/files#diff-43635fa581ff608d63be293b3806a4f3045dfaad79a9c48e1a739b90f177cc9a) file, it should be enough to prevent the issue you mentioned.

durera (changes requested):
Needs a section added to the README detailing the new `mas_special_characters ` variable and what it does (and if relavent which versions of MAS support it)

durera (dismissed):

PR #1414, opened 2024-08-15
[patch] Rename AI Broker role variables for consistency

karol-czarnecki:
This PR is related to https://github.com/ibm-mas/python-devops/pull/19

Also added condition to check if cluster has pipelines already installed. If not install if yes skip this step

Condition tested on:
- fyre cluster with cli install
- techzone with mas core 9.0 preinstalled - where pipelines already installed


durera (approved):
Approved, but I think we want to align the deliver of this with the other PRs in python-devops and CLI ...

- https://github.com/ibm-mas/python-devops/pull/19
- https://github.com/ibm-mas/cli/pull/1179

durera (approved):

PR #1418, opened 2024-08-16
[patch] Support OCP version aliases

IanBoden:
This adds two special OCP versions, "default" and "rotate".

default - uses the default_ocp_version as specified in the defaults, useful for long lived test/staging environments
rotate - uses an ocp_version based on the day of the week, useful for daily test environments

The versions only specify a minor number so will not work on aws where the patch needs to be specified


whitfiea (dismissed):
Can we update the readme with these two values https://github.com/ibm-mas/ansible-devops/tree/ib.ocpversion/ibm/mas_devops/roles/ocp_provision#ocp_version

durera (approved):

PR #1432, opened 2024-08-24
[patch] Add AIbroker public cert

zxue:
The MAS AI Broker public cert is missing. It is required when cert management is set to manual in OpenShift.


whitfiea:
@karol-czarnecki as this is AIBroker related can you review if this is correct?

karol-czarnecki (approved):
lgtm

whitfiea (approved):

PR #1476, opened 2024-09-18
[patch] Allow specifying CERT_MANAGER_NAMESPACE in suite_dns

andrercm:
Got two customer cases recently where IBM Certificate Manager and CIS webhook deployment has been wrongly deleted and in all these cases I had to reconfigure CIS webhook in `ibm-common-services` namespace even though cert manager was deployed as standalone, so in this case i need to be able to specify the `CERT_MANAGER_NAMESPACE` otherwise current code will automatically use `cert-manager` namespace and we don't want that for this particular case.


whitfiea (approved):

whitfiea (changes requested):
Should this be added to the readme https://github.com/ibm-mas/ansible-devops/blob/4a3b68e0a333abcd45923af5bf0b33d106f2dc98/ibm/mas_devops/roles/suite_dns/README.md ?

whitfiea (approved):

PR #1495, opened 2024-10-01
[patch] Switch Prometheus SA token creation from oc cli to ansible module

Delgon:
In order not to use the 'oc cli'. We switch the creation of the token of the prometheus service account via the creation of a dedicated secret and we retrieve it via the ansible k8s_info module.

Since Openshit 4.16, the creation of an SA no longer automatically generates a token. So we have to create it.


whitfiea:
@IanBoden @twhart can you review this PR as it has come in with the existing work related to support OCP 4.16.

twhart (approved):

twhart:
Looks good to me, I ran and tested everything and it works. Looks like creating the secret is also a better way to manage the expiry times on the tokens

whitfiea (approved):

PR #1535, opened 2024-10-30
[minor] Add Ansible Automation Platform Execution Environment

whitfiea:
Adds a new image built in this repo that extends the Red Hat Ansible Automation Platform(AAP) supported image. This allows users to use the ansible-devops-ee image as an execution environment in AAP.

It installs the same collection from galaxy that is built in this repo each time. It also installs a number of required cli and other libraries required to run.

This PR also includes the documentation on how to setup AAP with this execution environment and gives some examples and hints as to how to write your own playbooks to work in AAP.

I have tested this in my own AAP install, and this can be seen in the documentation screenshots.

Also fixed a few doc problems with missing role docs or incorrect links


durera (dismissed):
LGTM; Per discussion in Slack.

Issue #1346, opened 2024-06-26
COS role should use a targeted variable for IBM Cloud API Key


durera:
The COS role uses `IBMCOS_RESOURCEGROUP` to allow COS specific setting of the group, but does not do the same for the API Key, this means that we are forced to use an over-priviledged API key for the COS integration ... it should work the same way:

- cos_apikey
- cos_resourcegroup

The MAS tekton piptlines and CLI will need matching updates too.

durera:
https://jsw.ibm.com/browse/MASCORE-4345

PR #1555, opened 2024-11-13
[patch] Fixes and updates for IBMCloud COS support

twhart:
Setup COS so it doesn't rely on the ibmcloud_apikey if the cos_apikey has been set. This allows for a less privileged apikey to be used. Renamed some variables to align more closely to how the rest of COS is used

Tested here:
https://dashboard.masdev.wiotp.sl.hursley.ibm.com/tests/pfvtcos

See also: MASCORE-4345

### DIFF
--- a/ibm/mas_devops/roles/cos/templates/ibm/objectstoragecfg.yml.j2
+++ b/ibm/mas_devops/roles/cos/templates/ibm/objectstoragecfg.yml.j2
@@ -3,7 +3,7 @@ apiVersion: v1
 kind: Secret
 type: Opaque
 metadata:
-  name: cos-credentials-system
+  name: ibmcos-credentials-system
   namespace: "mas-{{ mas_instance_id }}-core"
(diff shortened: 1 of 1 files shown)


twhart:
Closes #1346

ibm/mas_devops/roles/cos_bucket/README.md
durera, line 1: I don't think we should rename this here, it will make this update incompatible with previous, and the instance name is fine to stay as "x-ibmcos-x"

ibm/mas_devops/roles/cos/templates/ibm/objectstoragecfg.yml.j2
durera, line 1: I think we should keep the same names here, it will avoid this becoming incompatible with previous runs, and there's no specific reason the name of the secret needs to change,

durera (changes requested):

durera (approved):

PR #1560, opened 2024-11-16
[patch] make cos_instance_name unique

lokesh-sreedhara:
https://jsw.ibm.com/browse/MASCORE-4379

Looking at fvt-release are we unable to run multiple in parallel 
Looks like they get upset about COS because all the releases use the same name

We should be fixing that so it uses a unique name, instead of assuming the mas instance id is unique across all runs

slack discussion: https://ibm-watson-iot.slack.com/archives/C031Q7W21FZ/p1730889780253499

### DIFF
--- a/ibm/mas_devops/roles/cos/tasks/providers/ibm/provision.yml
+++ b/ibm/mas_devops/roles/cos/tasks/providers/ibm/provision.yml
@@ -4,6 +4,12 @@
     that: cos_apikey is defined and cos_apikey != ""
     fail_msg: "cos_apikey property is required"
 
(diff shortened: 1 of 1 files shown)


durera (changes requested):
We don't want to change the default in the role ... we want to use the ability to manually set this in our FVT framework .. the default value we have is fine.

The change we need is that in our FVT framework we set `cos_instance_name = "Object Storage for MAS - {{ mas_instance_id }} {{ mas_channel }}"` for example, so that we generate names like fvtrelease 8.10.x, fvtrelease 8.11.x, and so on.

PR #1573, opened 2024-11-24
[major] Use new OLM functions in mas-devops

durera:
Remove the `install_operator` role & consolidate the different ways we manage IBM entitlement and create subscriptions in the Ansible collection ... now it is consistent, as below:

```
tasks:
  - name: "Update IBM Entitlement Key"
    ibm.mas_devops.update_ibm_entitlement:
      namespace: test-namespace
      icr_username: xxx
      icr_password: xxx
      artifactory_username: xxx  # Optional, for development builds
      artifactory_password: xxx  # Optional, for development builds
    register: secret

  - name: "Apply Subscription"
    ibm.mas_devops.apply_subscription:
      namespace: test-namespace
      package_name: ibm-sls
      package_channel: 3.x
    register: subscription
```

image


whitfiea (changes requested):
Should this not be major as we are removing the install_operator role?

PR #1574, opened 2024-11-25
[patch] Fix missing catalog_digest

terenceq:
## Failure with mirroring of the catalog

The catalog_digest is not being set when executing mirroring. The variable is available in  `mas_catalog_metadata.catalog_digest` and should be set when preparing the extra images.


andrercm (approved):
LGTM

PR #1610, opened 2024-12-23
[patch] no longer exclude events-operator

IanBoden:
Not sure what changed but we now need the events-operator to mirror common services


IanBoden:
Fixed under #1595

PR #1614, opened 2025-01-08
[patch] update dns entries for aibroker

roshi-dubey:
https://jsw.ibm.com/browse/MASAIB-494


whitfiea (changes requested):
I would think we need edge certificates as well? or is it to be treated as just a DNS entry with no proxy. https://github.com/ibm-mas/ansible-devops/blob/master/ibm/mas_devops/roles/suite_dns/templates/edge_certificate_routes.yml.j2

Also would need the entries here https://github.com/ibm-mas/ansible-devops/blob/5653ebdafa56e9cdf040a0ff82fb2e0dd3ed6d01/ibm/mas_devops/roles/suite_certs/templates/dnsentries.yml.j2#L53 if manual certs are supported.

karol-czarnecki (approved):
lgtm

PR #1631, opened 2025-01-28
 [patch] introduce filter var IMAGE_GROUP_FILTER to set --filter into image mirror

anilprajapatiibm:
MASCORE-5426

If Passed non-empty, non-false value to IMAGE_GROUP_FILTER, value will be pass filter as below
# --filter {{ image_group_filter }}


ibm/mas_devops/roles/mirror_case_prepare/defaults/main.yml
```diff
@@ -30,3 +30,7 @@ mirror_working_dir: "{{ lookup('env', 'MIRROR_WORKING_DIR') }}"
 ibmpak_skip_dependencies: "{{ lookup('env', 'IBMPAK_SKIP_DEPENDENCIES') | default('False', True) | bool  }}"
 ibmpak_skip_verify: "{{ lookup('env', 'IBMPAK_SKIP_VERIFY') | default('False', True) | bool  }}"
 ibmpak_insecure: "{{ lookup('env', 'IBMPAK_INSECURE') | default('False', True) | bool }}"
+
+# Below set the filter, if Passed non-empty, non-false value
+# --filter {{ image_group_filter }}
+image_group_filter: False
```
durera, line 36: This is defined as a boolean here, but as a string in the playbook.  

This also is not consistent with how we deal with env vars ... this role should have the env var lookup, thus making the addition of the `image_group_filter` to the playbook redundant.  e.g. see the other defaults here such as:

```
ibmpak_insecure: "{{ lookup('env', 'IBMPAK_INSECURE') | default('False', True) | bool }}"
```

Also, the implication here is that `image_group_filter`  is a global setting that we want to apply for all case bundles that are mirrored, rather than a setting that is per case bundle?  Is that correct, doesn't seem right to me.

durera (dismissed):

ibm/mas_devops/roles/mirror_case_prepare/defaults/main.yml
anilprajapatiibm: If we go with "{{ lookup('env', 'IBMPAK_INSECURE') | default('False', True) | bool }}" Instead of Default False,
this filter applies to all the mirror prepare command,

Questions I have:
1. Do you want me to introduce new Variable, Just for Db2u Filter?
2. Do you want me to add --filter db2 as default from Feb patch?

ankitsibm: @durera could you please resolve @anilprajapatiibm query & provide your inputs..  Fixed is done but as you have concern related to the env variable lookup its increasing dependency to make necessary changes further. Appreciate you quick feedback to resolve this conversation. Thanks!!

durera: Taken discussion to team channel in slack, easier

alequint (approved):

PR #1670, opened 2025-02-27
[patch] Increase wait time for application readiness check before wor…

anujkabraibm2022:
Added retry logic (15 attempts with 20s delay) before asserting application readiness Ensured `app_cr_result.resources` is defined and has at least one resource before checking status Prevents unnecessary failures due to delayed application startup in OpenShift Keeps `assert` after waiting to strictly validate application readiness because getting issue on

### DIFF
--- a/ibm/mas_devops/roles/suite_app_config/tasks/main.yml
+++ b/ibm/mas_devops/roles/suite_app_config/tasks/main.yml
@@ -94,7 +94,7 @@
     - app_cr_result.resources is defined
     - app_cr_result.resources | length > 0
     - app_cr_result.resources | json_query('[*].status.conditions[?type==`Ready`][].status') | select('match', 'True') | list | length == 1
-  failed_when: app_cr_result.failed
+  
 
 - name: "Check that the application is ready to configure a workspace"
   assert:
--- a/ibm/mas_devops/roles/suite_app_config/tasks/main.yml
+++ b/ibm/mas_devops/roles/suite_app_config/tasks/main.yml
@@ -94,7 +94,7 @@
     - app_cr_result.resources is defined
     - app_cr_result.resources | length > 0
     - app_cr_result.resources | json_query('[*].status.conditions[?type==`Ready`][].status') | select('match', 'True') | list | length == 1
-  failed_when: app_cr_result is failed
+  failed_when: app_cr_result.failed
 
 - name: "Check that the application is ready to configure a workspace"
   assert:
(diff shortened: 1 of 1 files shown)


whitfiea (commented):
Has this change been tested? I am not sure it will do what you think it will do as it isn't re getting the info.

Also can you update the PR description as it just says what is changed rather than why it was changed.

anujkabraibm2022:
What Changed? -
Removed the initial k8s_info lookup that ran only once.
Moved the k8s_info lookup inside the retry loop, so each retry fetches fresh application status.
Added a final assert check after retries to strictly validate application readiness before proceeding.
Why Was This Changed?
The test failed intermittently because the readiness check was not refreshing application status.
Previously, the retry loop checked stale data, leading to false failures when the application took longer to start.
Now, every retry fetches the latest status from OpenShift, ensuring the test waits properly instead of failing too soon.

ibm/mas_devops/roles/suite_app_config/tasks/main.yml
```diff
@@ -81,13 +81,24 @@
...
+- name: "Wait for application to be Ready before configuring workspace"
+  retries: 15  # Number of retries before failing
+  delay: 20    # Wait 20 seconds between retries
+  register: app_ready
+  until:
+    - app_ready.resources is defined
+    - app_ready.resources | length > 0
+    - app_ready.resources | json_query('[*].status.conditions[?type==`Ready`][].status') | select('match', 'True') | list | length == 1
+  failed_when: app_ready is failed
+  block:
```
whitfiea, line 93: why is this in a block? I am not sure this even works, has it been tested?

whitfiea (changes requested):

ibm/mas_devops/roles/suite_app_config/tasks/main.yml
```diff
@@ -81,13 +81,21 @@
...
     namespace: "{{ mas_app_namespace }}"
     kind: "{{ mas_app_kind }}"
   register: app_cr_result
+  retries: 15  # Number of retries before failing
+  delay: 20    # Wait 20 seconds between retries
+  until:
+    - app_cr_result.resources is defined
+    - app_cr_result.resources | length > 0
+    - app_cr_result.resources | json_query('[*].status.conditions[?type==`Ready`][].status') | select('match', 'True') | list | length == 1
+  failed_when: app_cr_result.failed
```
whitfiea, line 97: I don't think we need the failed_when, especially as we still have the assert. Has this been tested?

whitfiea (changes requested):

ibm/mas_devops/roles/suite_app_config/tasks/main.yml
anujkabraibm2022: @whitfiea removed please check

PR #1780, opened 2025-05-27
[minor] Support June Catalog Update

durera:
June catalog includes:
- MAS 9.1.0 GA
- Patch updates for all in-support releases
- PPC Catalog
- OCP 4.17 and 4.18 support
- CPD 5.1.3 support

See also:
- https://github.com/ibm-mas/cli/pull/1614
- https://github.com/ibm-mas/python-devops/pull/77


ibm/mas_devops/playbooks/mirror_add_iot.yml
durera, line 1: Why have we changed the order here, I ask because it shouldn't matter what order things are mirrored in.

ibm/mas_devops/playbooks/mirror_dependencies.yml
```diff
@@ -266,8 +266,8 @@
     - role: ibm.mas_devops.mirror_images
       when:
         - mirror_common_svcs or (cpd_48_or_higher and mirror_cp4d)
-        - mas_catalog_metadata.common_svcs_version_1 is defined
```
durera, line 269: Why are we changing the alternative common services version variable from `_1` to `_2`, this is going to break anything using the old variable, and seems to serve no purpose - we're not adding a third version to mirror, we're just swapping a variable name.

IanBoden: we already did _1 on line 248, this is for mirroring _2 so in the situation where _1 is defined and _2 isn't this breaks. It needs to only try to mirror _2 if _2 is defined. In this catalog we only need 2 versions (common_svcs_version and common_svcs_version_1)

ibm/mas_devops/common_vars/compatibility_matrix.yml
```diff
@@ -46,45 +55,55 @@ compatibility_matrix:
...
+    9.1.x: [9.0.x]
+    9.1.x-feature: [9.0.x]
+    9.0.x: [8.9.x]
+    8.9.x: [8.8.x]
+    8.8.x: [8.7.x]
 
 # There is probably a better way to do this by manipulating the channel string, but this is fast and cheap!
 upgrade_path:
-  9.0.x: 9.1.x-feature
+  9.1.x: 9.1.x # To handle when core is already on 9.1.x but Apps needs upgrade
```
anilprajapatiibm, line 104: @durera  can you please once review this change?

PR #1781, opened 2025-05-28
[patch] Add OCP 4.18 into rotation

durera:
- Improves error handling when provisioning FYRE clusters with an invalid/expired API key
- Adds OCP 4.18 into the `rotate` policy for `ocp_provision` role


durera:
> will this break fvt's that run in ibm cloud or have they added 4.18 support now?

Yes, 4.18 is available in IBM Cloud

PR #1798, opened 2025-06-09
[patch] Adding ODF dependencies for mirroring in OCP

jainyjoseph:
Adding ODF dependencies for mirroring in OCP
-----------------------------------------------
Updated https://github.com/ibm-mas/ansible-devops/blob/master/ibm/mas_devops/roles/mirror_ocp/templates/imagesetconfiguration.yml.j2#L87 , to add the ODF dependencies for redhat-operator catalogs

Tested same in a fyre airgap test cluster and found ODF gets installed successfully.


durera:
These changes are already in master, closing PR

PR #1802, opened 2025-06-12
[patch] fix JDBC secret creation for aibroker role

boonware:
## Issue
MASAIB-1028

## Description
* Ensure that the "SSL Enabled" option has a default boolean value when read from file or environment. We can then eliminate the validation step that checks the length of the boolean, which throws an error for obvious reasons.
* Base64 encode all values directly in the secret, rather than selectively encoding some of them in task, which is confusing. This involves decoding two secret values first, and adding a comment to make it clear where specific values are being read from.

<hr/>


karol-czarnecki (changes requested):
could you please remove those lines https://github.com/ibm-mas/ansible-devops/blob/master/ibm/mas_devops/playbooks/oneclick_add_aibroker.yml#L36-L45

karol-czarnecki (approved):

whitfiea (approved):

PR #1821, opened 2025-07-02
[patch] Add missing var in aibroker-tenant role

Bhautik-Vala:
## Issue

Add missing var in aibroker-tenant role. before it was read from oneclick playbook but for CLI purpose we have to add in aibroker-tenant role also.
## Test Results

<hr/>


jasmin-30 (approved):
LGTM

whitfiea (changes requested):
This shouldn't be needed as the aibroker role will install the operator

Bhautik-Vala:
> This shouldn't be needed as the aibroker role will install the operator

Yes you are right, but right now we have issue with encoding of certs in workspace operator.. and we conditionally encode the certs here - https://github.com/ibm-mas/ansible-devops/blob/master/ibm/mas_devops/roles/aibroker_tenant/templates/aibroker/aibrokerworkspace.yml.j2#L26
so that's why we need mas_app_channel as well

PR #1822, opened 2025-07-04
[patch] Don't set SLS domain by default

rawa-resul:
## Test Results

<hr/>

### DIFF
--- a/ibm/mas_devops/roles/sls/tasks/gencfg/main.yml
+++ b/ibm/mas_devops/roles/sls/tasks/gencfg/main.yml
@@ -2,47 +2,62 @@
 ###############################################################################
 #                             New / Existing SLS                              #
 ###############################################################################
+- name: Obtain sls-suite-registration ConfigMap
+  when: (sls_tls_crt is not defined or sls_tls_crt == "") or
+        (sls_registration_key is not defined or sls_registration_key == "") or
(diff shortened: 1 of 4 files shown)


rawa-resul:
> Did we set up existing MAS instances to use the route that we are now effectively deprecating (in most scenario)?
> 
> I assume the result of this change is that the slscfg that will be used in most installs will now be one that is using the kubernetes service.
> 
> What is the expected behaviour in this scenario:
> 
> * Install first MAS instance + SLS using older version of the CLI
> * Install second MAS instance using new version of the CLI
> 
> When the second install runs it will refresh the SLS instance ... and presumably this will result in the removal of the route. So did we set up the first MAS instance using that route?

When you install using an older version of CLI, SLS will have a Route. If you then install over it using a newer CLI with this change then nothing will happen to the Route. The domain field will still be present in the LicenseService CR. This is the behaviour I was aiming for. I did not want to break anyone's environment if they already have a Route because they could be using it off-cluster too. 

When patching the LicenseService CR with a spec that doesn't have the domain field would not remove it. By default, Kubernetes only updates fields that are in both the local template and CR, and adds the remaining new fields.

Moreover, SLS does not currently have the logic of removing a Route when the domain field is removed from the CR. This was discovered during this work and I will add that into SLS.

durera (approved):

PR #1880, opened 2025-08-12
[patch] split idms for aiservice

IanBoden:
## Issue

MASCORE-8455

## Description

Split ImageDigestMirrorSet for aiservice and mas
Added ImageTagMirrorSet for aiservice
Changed to look up subscription info from the package manifest to allow installing subscriptions with non default catalog source names (in airgap we don't use the default names)

## Test Results
Runs in an airgap environment (we don't get all the way through but far enough to show this works)


durera:
Closing PR due to old age

PR #1886, opened 2025-08-17
[patch] fix db2u facilties install

anilprajapatiibm:
## Issue
without this change facilities db2 setup failing with db2u latest version: `v110509.0.5`

Below steps we did to validate the changes:
1. Manually delete the db2u pod
2. Wait for new pod to come up and ready
3. wait for jdbccfg to reconcile, it was working fine after recreating the pod

It means somewhere in the previous or current db2u version issue: `that was causing TLS connection issues when the pod was first started` have been resolved

so removing the code:
```
# 16. Delete db2 pod on intial setup only, to workaround db2 issue
# that was causing TLS connection issues when the pod was first startedà
# Comment out becuase the database could end-up in an inconsistent state
- name: Delete db2u pod on initial setup
  kubernetes.core.k8s:
    api_version: v1
    kind: Pod
    name: "c-{{db2_instance_name | lower}}-db2u-0"
    namespace: "{{db2_namespace}}"
    state: absent
  when: initial_db2_cluster_lookup.resources | length == 0
```

## Test Results
Tested on pfvt:
https://dashboard.masdev.wiotp.sl.hursley.ibm.com/tests/pfvtdb2u

image

<hr/>

### DIFF
--- a/ibm/mas_devops/roles/db2/tasks/install/main.yml
+++ b/ibm/mas_devops/roles/db2/tasks/install/main.yml
@@ -371,19 +371,8 @@
     namespace: "{{db2_namespace}}"
     state: absent
 
-# 16. Delete db2 pod on intial setup only, to workaround db2 issue
-# that was causing TLS connection issues when the pod was first started
-# -----------------------------------------------------------------------------
-# - name: Delete db2u pod on initial setup
-#   kubernetes.core.k8s:
-#     api_version: v1
-#     kind: Pod
-#     name: "c-{{db2_instance_name | lower}}-db2u-0"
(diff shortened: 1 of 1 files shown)


anilprajapatiibm:
Hi @terc1997 As you can see in the attached screenshot of the pfvt test,
I have tested it almost 9 times, and it was working fine on all the test specific to manage db2u installation and setup

AntonioPerrone:
Hi @terc1997, I have performed multiple test tests and I have not observed any issue with TLS. We need to proceed with this and take under control it. If we observe any issue that would require the DB2 pod to be restarted, we need to do that in a less disruptive way respect to just killing the POD.

PR #1904, opened 2025-08-26
[patch] Add ODH Model deployment type option and set it to raw as default

jasmin-30:


durera (changes requested):
Changes discussed in Slack: https://ibm-watson-iot.slack.com/archives/C096J1Q1U3H/p1756241208909789

PR #1909, opened 2025-08-28
[minor] Fix for issue reported in roks_provisioning in the case where a cluster already exists

jainyjoseph:
## Issue Description

There is a issue reported in MAS performance automation pipeline for ocp_provision_roks

Screenshot 2025-08-28 at 7 57 33 AM

## Description
The issue is happening because when cluster already exists, it is  not handled at Parse VLAN list task.

## Test Results
- Able to  reproduce the issue
   
Screenshot 2025-08-28 at 1 51 37 PM

- Fix applied so that it handles the case when cluster already exists and the task is skipped
Screenshot 2025-08-28 at 1 56 21 PM

- Other scenerios re-tested:
  Case 1 : Private and public vlan doesnot exist
  Case 2: Private vlan exist, public vlan is missing
  Case 3: Both private and public vlan exist
  Case 4: Public vlan exist and private vlan doesnot exist
  
  All these 4 cases working as expected, current change nowhere disturbing these scenerios.


terc1997 (approved):
LGTM

PR #1937, opened 2025-10-01
[patch] Add to AI Service support for watsonX CA cert

karol-czarnecki:
## Issue
https://jsw.ibm.com/browse/MASAIB-1432

## Description
Add support to use WastonX AI CA certificate

## Test Results
Deployment pass in both cases - with WatsonX AI CA, without WatsonX AICA


ibm/mas_devops/roles/aiservice/tasks/config_wx/main.yml
```diff
@@ -0,0 +1,15 @@
+---
+- name: Delete WatsonX secret if exists
+  kubernetes.core.k8s:
+    api_version: v1
+    kind: Secret
+    name: "{{ aiservice_wx_cacert_secret }}"
+    state: absent
+    namespace: aiservice-{{ aiservice_instance_id }}
+  ignore_errors: true
```
whitfiea, line 9: why does the secret need to be deleted? the apply should ensure that the secret has the correct values

karol-czarnecki: fixed this issue - thank you

PR #2019, opened 2025-12-08
[patch] Fix check for AIService upgrade compatibility

boonware:
## Issue
MASAIB-1626

## Description
* To determine if the upgrade path is valid the code currently performs the check `aiservice_channel in aiservice_upgrade_path`, however, `aiservice_upgrade_path` is a dictionary so this check is insufficient; we need to check if (a) the currently installed AI Service channel is a valid key in the dictionary, and (b) if the values for this key include the channel version we want to upgrade to. The PR adds a new filter function to perform this check.
* Another filter function is added to get the default target channel from the dictionary when a user does not supply one. **The first entry in a list of target versions is taken as the default.**
* Includes unit tests for the new filter functions. Even though the pipeline does not currently run these tests they are useful locally for a developer to run.

## Test Results
AI Service upgrade has been successfully tested with these changes included in custom branch https://github.com/ibm-mas/ansible-devops/pull/2021

<hr/>

### DIFF
--- a/ibm/mas_devops/plugins/filter/filters.py
+++ b/ibm/mas_devops/plugins/filter/filters.py
@@ -1,10 +1,3 @@
-# -----------------------------------------------------------
-# Licensed Materials - Property of IBM
-# 5737-M66, 5900-AAA
-# (C) Copyright IBM Corp. 2021 All Rights Reserved.
-# US Government Users Restricted Rights - Use, duplication, or disclosure
-# restricted by GSA ADP Schedule Contract with IBM Corp.
-# -----------------------------------------------------------
 import yaml
 import re
 


ibm/mas_devops/plugins/filter/test_filters.py
```diff
@@ -0,0 +1,70 @@
+# -----------------------------------------------------------
```
durera: Errrrrrm, I hope this is just a mistake of putting the wrong license file in place, rather than copying product code into this repo? 

ansible-devops is open source code, licensed under the EPL, we can not published licensed materials in this repository

durera (approved):

PR #2073, opened 2026-01-21
[minor] B&R for catalog, certmanager, db2, mongo

sanjayprab:
## Description

PR for dev->dev branch, raised for review purpose.

B&R for 
- catalog
- cert-manager
- db2
- mongo

Reverted  "install-from-backup" changes
Removed br_skip_instance flag
Added backup_database action for mongodb and db2
Updated docs.

## Test Results
tested locally on br test

### DIFF
--- a/ibm/mas_devops/roles/mongodb/tasks/main.yml
+++ b/ibm/mas_devops/roles/mongodb/tasks/main.yml
@@ -10,8 +10,8 @@
   assert:
     that:
       - mongodb_action is defined
-      - mongodb_action in ["install", "uninstall", "backup", "restore_database", "restore"]
-    fail_msg: "mongodb_action property is required and must be one of 'install', 'uninstall', 'backup', 'restore_database', 'restore'"
+      - mongodb_action in ["install", "uninstall", "backup", "backup_database", "restore_database", "restore"]
+    fail_msg: "mongodb_action property is required and must be one of 'install', 'uninstall', 'backup', 'backup_database','restore_database', 'restore'"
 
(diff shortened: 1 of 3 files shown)


ibm/mas_devops/roles/db2/tasks/restore/determine-storage-classes.yml
```diff
@@ -0,0 +1,15 @@
...
+# Get Storage from DB2 CR
+- name: Override storage class names in the db2 spec.storage
+  set_fact:
+    _db2_storages: "{{ db2ucluster_cr_cfg.spec.storage | ibm.mas_devops.override_db2_storage_classes_names(db2_meta_storage_class, db2_data_storage_class, db2_backup_storage_class, db2_logs_storage_class, db2_temp_storage_class) }}"
+  when:
+    - db2_meta_storage_class is defined
+    - db2_data_storage_class is defined
+    - db2_backup_storage_class is defined
+    - db2_logs_storage_class is defined
+    - db2_temp_storage_class is defined
```
whitfiea: does this mean we only override if all are set to override? What happens if we don't have db2 temp storage?

whitfiea (changes requested):

ibm/mas_devops/roles/db2/tasks/restore/determine-storage-classes.yml
sanjayprab: Yes it wont set it if the storage is not present in the CR.

PR #2093, opened 2026-01-29
[minor] Introduce safe CP4D uninstall role

harsh-patel14:
## **Description:**
This PR introduces a **robust and safe uninstall workflow** for **Cloud Pak for Data (CP4D)**.

The changes focus on:
- **Making CP4D uninstall safe by default** with explicit user confirmation.
- **Handling partially uninstalled and broken states** commonly seen in real clusters.  
- **Removing operator- and service-level finalizers** that block namespace deletion.  

## **Test Results:**
Validated against a **OpenShift cluster** with:
- **CP4D 5.2.**
- **WML, WSL, CA, and SPARK** services installed.

### **Tested scenarios:**
- **Full CP4D uninstall.**
- **Namespaces stuck in `Terminating` state.**
- **Remaining service CRs with finalizers after operator removal.**

### **Verified outcomes:**
- **CP4D presence is detected before any uninstall actions.**
- **Uninstall safely exits when not approved.**
- **Finalizers are removed only for CP4D-related resources.**
- **Namespaces are successfully deleted once blocking resources are cleaned.**

### **Other notes:**
- Internal Jira: mascore-3742.
- I have attached log file of uninstall process for your varification purpose.
- I would be like to have more suggestions to enhnace or rectify anything this role includes.


ibm/mas_devops/roles/cp4d_uninstall/meta/main.yml
```diff
@@ -0,0 +1,35 @@
+#SPDX-License-Identifier: MIT-0
+galaxy_info:
+  author: your name
```
durera: This file is still generated boilerplate

durera (changes requested):

terc1997:
Include upstream.

PR #2140, opened 2026-03-04
[minor] Support db2 v12 upgrade

terc1997:
## Description
To support Db2 v12 upgrade, it was required to change the update task in db2 requiring approval of the upgrade by the user leveraging the variable `db2_v12_upgrade`. Also, Db2 v12 requires an activation key that is delivered in form of license file. This has been introduced into the install mode through `db2_license_file` and the env var `DB2_LICENSE_FILE`. Also, the configuration that it's set for Manage for `STMTHEAP` is too small to perform the upgrade in Manage. Therefore, it was added a patch to set the `STMTHEAP` to `20000 AUTOMATIC`. So, for the support will be included:
- `validate_upgrade` task to verify if it will be performed a major upgrade of Db2.
- `db2_v12_upgrade` variable to ensure user approval before upgrading. 
- `db2_license_file` variable with the path of the license
- `patch-smtmheap.yml` patches the configuration in the Db2 to increase the heap size for Db2 statement bigger than 20000 bytes. These change has been also included for future deployments of Db2 for Manage.

### Install with License
image

### FVT Tests
 - db2upgrade3
 - db2instance
<hr/>


durera (commented):
We need a readme update with this that covers v11 -> v12 upgrade and v12 license file, including - I think what would be useful - an example of what a db2 license file looks like, so that a customer can compare what they think they have is actually what we're asking for, because IBM loves to call things by multiple names ... the sample shouldn't be a real license file, but can be used to check "is this thing I have, that I might call by a different name, the thing you want me to use here".

terc1997:
Added documentation and made some changes to guarantee idempotency to the scripts.

durera (approved):

PR #2188, opened 2026-04-09
[patch] improve checks for installing AI Service 9.2 tenant operator

boonware:
## Description
* When the channel is not 9.1 we currently install the new AI Service namespace-scoped tenant operator. This will cause a failure if the current feature channel release is installed since the catalog does not yet contain a feature channel release of this operator. To solve this we use the "aiservice_migration" role to perform a check which determines if the new operator is needed or not, and only attempt to install the operator if true. This facilitates current test efforts for the upcoming 9.2 release.
* Ensure the AI Service migration role automatically sets the same channel for the `ibm-aiservice-tenant` subscription based on the subscription for `ibm-aiservice`, with an allowed override for dev purposes.

## Test Results
Verified by installing the 9.2.x-feature channel and confirming that resources were installed to correct namespaces and that the 9.2 tenant operator was not installed.

<hr/>


joshiricha444 (dismissed):
LGTM

boonware:
Replaced by https://github.com/ibm-mas/ansible-devops/pull/2200

PR #2209, opened 2026-04-16
[patch] Temporarily add --remove-signatures to oc mirror

prveenkumr:
During mirror-image `gpu-operator, turbonomic, ibm-data-reporter, ibm-metrics-operator` image failing to mirror with error: 

 `reading signatures: reading manifest sha256-[digest].sig in registry.connect.redhat.com/[path]: name unknown: Image not found`

The mirror process is trying to read image signatures (.sig files) from registry.connect.redhat.com, but these signature manifests do not exist in the registry.

fix: 
Include the `--remove-signatures` option in oc mirror invocations across mirror_ocp tasks. Updated direct.yml, to-filesystem.yml, and from-filesystem.yml so mirrors (direct, to filesystem, and from filesystem) strip image signatures during the mirror operations; other command arguments and log redirections remain unchanged.

**We need to get Red Hat to deliver a fix to their certified operator catalog so that image mirroring works again**, we will revert this change as soon as that is done.

<hr/>

### DIFF
--- a/ibm/mas_devops/roles/mirror_ocp/tasks/actions/direct.yml
+++ b/ibm/mas_devops/roles/mirror_ocp/tasks/actions/direct.yml
@@ -61,10 +61,12 @@
       ignore_errors: true
       loop: "{{ repos_to_add }}"
 
+# Important: We have added "--remove-signatures" temporarily due to a problem in the Red Hat Certified operator catalog
+# This is the only way we can get image mirroring working.  We will remove this as soon as Red Hat fix the operator catalog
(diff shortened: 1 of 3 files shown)


IanBoden:
As discussed we can't do this for all images we need a different solution

durera:
I've looked for less sledgehammer solutions, there aren't any I can see.  Ultimately Red Hat need to fix their certified operator catalog, but until they do, this solution at least unblocks image mirroring.

durera (approved):

PR #2231, opened 2026-05-04
[patch] Add condition to check for IPv6 CIDR in login-fyre.yml

JigarPatel2691:
## Description
to overcome the duplicate entries in network configuration, where we use existing cluster

## Test Results
tested by running the run role

image

now it's skipping this changes and so we will not have duplicate entries of ipv6 cidrs
<hr/>

### DIFF
--- a/ibm/mas_devops/roles/ocp_login/tasks/login-fyre.yml
+++ b/ibm/mas_devops/roles/ocp_login/tasks/login-fyre.yml
@@ -74,4 +74,4 @@
   register: _dualstackNetworkAddr
   when:
     - enable_ipv6 == true
-    - "'fd01::/48' not in (_network.resources[0].spec.clusterNetwork | map(attribute='cidr') | list)"
+    - "'fd01::/48' not in (_network.result.spec.clusterNetwork | map(attribute='cidr') | list)"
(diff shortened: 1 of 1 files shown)


durera (changes requested):
This "Enable dual stack networking" task should not even be part of the `ocp_login` role.  We have [`ocp_config`](https://github.com/ibm-mas/ansible-devops/tree/master/ibm/mas_devops/roles/ocp_config) for various post provision configuration tasks which are enabled by different env vars, this should be happening in that, rather than as part of the login.

JigarPatel2691:
> This "Enable dual stack networking" task should not even be part of the `ocp_login` role. We have [`ocp_config`](https://github.com/ibm-mas/ansible-devops/tree/master/ibm/mas_devops/roles/ocp_config) for various post provision configuration tasks which are enabled by different env vars, this should be happening in that, rather than as part of the login.

Agree, Thank you @durera for navigating to the right direction. Iet me cancel this PR, I will create seperate PR after testing it to the personal fvt.

JigarPatel2691:
will raise seperate PR, after removing ipv6 config code from login to config, and will test in pfvt environment

PR #2233, opened 2026-05-06
[patch] update the edge certificate routes, console url for path mode

dixitgsathwara:
## Description

- update the edge certificate routes to support path mode
- remove the default value for routingMode and ingressController because ansible-devops is a general repo used across all MAS versions, and path mode support is for 9.2+. If a value is defined, then only we need to consider it so that older MAS versions do not get this routingMode value. Right now, if we are on any version, this routingMode is set in the Suite CR spec.settings
 - update the console link 
## Test Results

<hr/>


IanBoden (approved):
Approved based on testing done was:
for the edgeconfig changes i have tested in my local by providing different vars flags to validate it generates the right yml file...
and where when i remove the routingMode value as it's required to not take the default one and i validated all the conditions where routingMode is undefined then fallback to subdomain.

PR #2276, opened 2026-05-26
[patch] Refactor Suite install template to group metrics under spec.settings.metricCollection

Hita-Pandya:
## Description
update the Suite install template in ansible-devops to render metric-related settings under spec.settings.metricCollection

## Test Results
validated the change by creating a new MAS instance.
image

### DIFF
--- a/ibm/mas_devops/roles/suite_install/templates/core_v1_suite.yml.j2
+++ b/ibm/mas_devops/roles/suite_install/templates/core_v1_suite.yml.j2
@@ -88,14 +88,15 @@ spec:
       allowSpecialChars: {{ mas_special_characters | bool}}
 {% endif %}
 {% if mas_channel != '8.7.x' and mas_channel != '8.8.x' and mas_channel != '8.9.x' and mas_channel != '8.10.x' and mas_channel != '8.11.x' and mas_channel != '9.0.x' %}
+    metricCollection:
 {% if mas_deployment_progression is defined and mas_deployment_progression != '' %}
-    deploymentProgression: {{ mas_deployment_progression | bool }}
+      deploymentProgression: {{ mas_deployment_progression | bool }}
 {% endif %}
 {% if mas_usability_metrics is defined and mas_usability_metrics != '' %}
-    usability: {{ mas_usability_metrics | bool }}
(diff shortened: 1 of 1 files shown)


ibm/mas_devops/roles/suite_install/templates/core_v1_suite.yml.j2
```diff
@@ -88,14 +88,15 @@ spec:
       allowSpecialChars: {{ mas_special_characters | bool}}
 {% endif %}
 {% if mas_channel != '8.7.x' and mas_channel != '8.8.x' and mas_channel != '8.9.x' and mas_channel != '8.10.x' and mas_channel != '8.11.x' and mas_channel != '9.0.x' %}
+    metricCollection:
```
durera, line 92: This should only be added to the CR if the user explicitly passes in env vars to configure these settings.  Let's not add things to the CR when there's no specific need to.

e.g. each var can have three expected values: true|false|anythingelse

`MAS_METRIC_COLLECTION_DEPLOYMENTPROGRESSION` = "true" or "false" means user made a choice
`MAS_METRIC_COLLECTION_DEPLOYMENTPROGRESSION` = "anything else" it means user doesn't care, we won't set this in the spec, and the operator will use it's default

If `MAS_METRIC_COLLECTION_DEPLOYMENTPROGRESSION`, `MAS_METRIC_COLLECTION_USABILITY`, `MAS_METRIC_COLLECTION_FEATUREUSAGE` are all "" then we don't need to add this to the spec at all.

durera (dismissed):

ibm/mas_devops/roles/suite_install/templates/core_v1_suite.yml.j2
lokesh-sreedhara: @durera the PI compliance and Opt Out feature design calls for this behaviour. We need the flags set to True by Default. 

So that the metering data flows from MAS to `segment` OOTB. We are providing the customer a setting to Turn this off if he wishes. 

Without the True value, we wont have metering data down stream on amplitude when Sales and PM's are gathering Customer Spend

sanjayprab (approved):

PR #2330, opened 2026-06-18
[patch] Fix bug for AI Service tenant upgrade

jasmin-30:
## Description

- When AIServiceTenant subscriptions is created, it doesn't have label `aiservice.ibm.com/instanceId = {{ aiservice_instance_id }}` because of which while running upgrade, tenant operators were not being selected for upgrade which may cause incompatibility with AI Service 9.2.x operator. 
- To address this issue, the label selector was removed from the subscription lookup, allowing all tenant subscriptions to be selected for upgrade.
- There was a bug which recently surfaced when upgrading tenant operator to 9.2.x, AIServiceTenant CR lookup was wrongly configured to use subscription name. Fixed the bug by using the namespace from subscription context which is essentially same for AIServiceTenant CR.

## Test Results

- Tested successful upgrade from 9.2.x-feature to 9.2.x.
- Tested successful upgrade from 9.1.x to 9.2.x

<hr/>

### DIFF
--- a/ibm/mas_devops/common_vars/compatibility_matrix.yml
+++ b/ibm/mas_devops/common_vars/compatibility_matrix.yml
@@ -129,4 +129,5 @@ upgrade_path:
 # The key is the current installed channel of AI Service
 # and the value is the allowed upgrade versions.
 aiservice_upgrade_path:
+  9.2.x-feature: [9.2.x]
   9.1.x: [9.2.x-feature]
--- a/ibm/mas_devops/common_vars/compatibility_matrix.yml
+++ b/ibm/mas_devops/common_vars/compatibility_matrix.yml
@@ -129,5 +129,4 @@ upgrade_path:
 # The key is the current installed channel of AI Service
 # and the value is the allowed upgrade versions.
 aiservice_upgrade_path:
-  9.2.x-feature: [9.2.x]
-  9.1.x: [9.2.x]
+  9.1.x: [9.2.x-feature]
--- a/ibm/mas_devops/common_vars/compatibility_matrix.yml
+++ b/ibm/mas_devops/common_vars/compatibility_matrix.yml
@@ -130,4 +130,4 @@ upgrade_path:
 # and the value is the allowed upgrade versions.
 aiservice_upgrade_path:
   9.2.x-feature: [9.2.x]
-  9.1.x: [9.2.x-feature]
+  9.1.x: [9.2.x]
--- a/ibm/mas_devops/roles/aiservice_upgrade/tasks/tenant/upgrade.yml
+++ b/ibm/mas_devops/roles/aiservice_upgrade/tasks/tenant/upgrade.yml
@@ -73,7 +73,7 @@
 - name: "Confirm AIServiceTenant CR is ready"
   kubernetes.core.k8s_info:
     api_version: aiservice.ibm.com/v1
-    name: "{{ tenant_subscription.metadata.name }}"
+    name: "{{ tenant_subscription.metadata.namespace }}"
     namespace: "{{ tenant_subscription.metadata.namespace }}"
     kind: AIServiceTenant
   retries: 20 # about 40 minutes
--- a/ibm/mas_devops/roles/aiservice_upgrade/tasks/upgrade_aiservice_tenant.yml
+++ b/ibm/mas_devops/roles/aiservice_upgrade/tasks/upgrade_aiservice_tenant.yml
@@ -11,19 +11,21 @@
 # If migration is not needed then proceed with simply upgrading the tenant operator
 - when: not is_migration_needed
   block:
-    - name: Find all subscriptions for ibm-aiservice-tenant
+    - name: Get all subscriptions
       kubernetes.core.k8s_info:
         api_version: operators.coreos.com/v1alpha1
         kind: Subscription
-        name: ibm-aiservice-tenant
-        label_selectors:
-          - "aiservice.ibm.com/instanceId = {{ aiservice_instance_id }}"
-      register: aiservice_tenant_sub_info
+      register: all_subscriptions
       no_log: true
 
-    - name: Set target resources
+    - name: Filter subscriptions named ibm-aiservice-tenant
       ansible.builtin.set_fact:
-        tenant_subscriptions: "{{ aiservice_tenant_sub_info.resources }}"
+        tenant_subscriptions: >-
+          {{
+            all_subscriptions.resources
+            | selectattr('metadata.name', 'equalto', 'ibm-aiservice-tenant')
+            | list
+          }}
 
     - name: Display subscription info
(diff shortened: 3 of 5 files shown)


ibm/mas_devops/common_vars/compatibility_matrix.yml
```diff
@@ -129,4 +129,5 @@ upgrade_path:
 # The key is the current installed channel of AI Service
 # and the value is the allowed upgrade versions.
 aiservice_upgrade_path:
-  9.1.x: [9.2.x-feature]
+  9.2.x-feature: [9.2.x]
+  9.1.x: [9.2.x]
```
IanBoden: should we be updating the compatibility_matrix in master for 9.2.x? shouldn't that go in the monthlyupdate branch

ibm/mas_devops/roles/aiservice_upgrade/tasks/tenant/upgrade.yml
```diff
@@ -73,7 +73,7 @@
 - name: "Confirm AIServiceTenant CR is ready"
   kubernetes.core.k8s_info:
     api_version: aiservice.ibm.com/v1
-    name: "{{ tenant_subscription.metadata.name }}"
+    name: "{{ tenant_subscription.metadata.namespace }}"
```
IanBoden, line 76: this doesn't look good, why are we setting the name to be the namespace?

ibm/mas_devops/roles/aiservice_upgrade/tasks/upgrade_aiservice_tenant.yml
```diff
@@ -11,19 +11,21 @@
...
-    - name: Set target resources
+    - name: Filter subscriptions named ibm-aiservice-tenant
       ansible.builtin.set_fact:
-        tenant_subscriptions: "{{ aiservice_tenant_sub_info.resources }}"
+        tenant_subscriptions: >-
+          {{
+            all_subscriptions.resources
+            | selectattr('metadata.name', 'equalto', 'ibm-aiservice-tenant')
+            | list
+          }}
```
IanBoden, line 28: not sure I see a benefit in doing it this way over using the labels which is the typical way to filter. If the label is missing why isn't the fix to add the label?

IanBoden (changes requested):
I'm not sure this change makes sense and merging it to master seems wrong too

ibm/mas_devops/roles/aiservice_upgrade/tasks/tenant/upgrade.yml
boonware: @IanBoden The AIServiceTenant CR always has the same name as it's namespace, by convention.

ibm/mas_devops/roles/aiservice_upgrade/tasks/upgrade_aiservice_tenant.yml
boonware: @IanBoden The "apply subscription" function does not support adding labels. To add labelling support would require updating Ansible and also the "mas-devops" library, which would required further extensive testing. The choice was made to go with this approach for now as this needs to be fixed urgently for the release, then afterwards add labelling support and improve this code.

boonware (dismissed):
Replied to some existing comments to give my take.

ibm/mas_devops/common_vars/compatibility_matrix.yml
jasmin-30: It's already updated in monthlyupdate branch.
Since the monthlyupdate branch will be merged to master branch eventually, I will remove the changes from this file.

ibm/mas_devops/roles/aiservice_upgrade/tasks/tenant/upgrade.yml
IanBoden: relying on a convention isn't good, if there is a policed rule that they are the same then maybe we would go with this but what happens when that changes? I don't see the benefit in tying this ability to a convention, what is wrong with using name?

jasmin-30: For 9.2.x versions, the name of the AIServiceTenant CR will always be the same as the namespace in which it is created.
If it changes in future, we will certainly revisit this implementation.

Ref: https://github.com/ibm-mas/ansible-devops/blob/master/ibm/mas_devops/roles/aiservice_tenant/templates/aiservice/aiservicetenant.yml.j2#L5-L10

IanBoden: but I don't understand why we would do something that needs to be revisited, why is this change needed?

ibm/mas_devops/roles/aiservice_upgrade/tasks/upgrade_aiservice_tenant.yml
IanBoden: can you point me to the issue that will address this (send via slack as you shouldn't post links to jira issues in public repos)

jasmin-30: Shared via slack

ibm/mas_devops/roles/aiservice_upgrade/tasks/tenant/upgrade.yml
jasmin-30: By design it is implemented in a way that the name and namespace will always be same for AIServiceTenant. 

> why is this change needed?

It was a defect left earlier due to which CR lookup was failing. This fixes the error.

IanBoden: so if the name and the namespace are the same why are you using the namespace and not the name?

jasmin-30: In AIService 9.2.x, the AIServiceTenant name and namespace are effectively the same identifier by design, and the namespace is the value we already have available from the subscription context. Using it fixes the lookup failure without introducing additional resource queries.

IanBoden: I'm still failing to see a justification of why this change is necessary. What is wrong with using the name?

jasmin-30: Subscription name is different than CR name. We are using subscription context to lookup Tenant CR.

jasmin-30: Discussed the purpose of this change over call and clarified the concerns.

IanBoden (approved):

PR #2347, opened 2026-06-30
[minor] Support ROSA HCP in ocp_idms role

rawa-resul:
## Description

Add support for setting up IDMS in a ROSA HCP cluster. This requires us to run ocp_config role first to add allowed registries and then run the ocp_idms role. If we don't add the allowed registries then we cannot add the idms. This is a bug and a workaround at the moment from RedHat's side. 

Changes:
- Move current `ocp_idms` role as `default` under its `tasks` directory
- Add new `rosa_hcp` approach under tasks in `ocp_idms` role
- Add `allowed registries` function in `ocp_config` role
  - This adds mas and redhat registries so that their IDMS can work 
- Add `additional trusted ca` function in `ocp_config` role
  - This adds the ca of the custom image registry
- Use rosa cli under the hood for some of the tasks. We need rosa cli `1.2.61` or greater for above to work.

## Related

- https://github.com/ibm-mas/cli-base/pull/33

## Test Results

<hr/>

Extensive manual testing was done in a ROSA HCP cluster and verified by expert labs as well.


durera (changes requested):
Think the approach in general is sound, but needs some tweaking around the use of CLUSTER_TYPE ... use of the same env var across all roles needs to be consistent, so I would suggestion the logic here should be:

`if CLUSTER_TYPE=rosa-hcp`: <do the special case>, otherwise <do the normal thing>

Otherwise when CLUSTER_TYPE=roks/fyre/ipi etc we will have broken the `ocp_idms` role.

As proposed above, we will do the "normal" way unless it's rosa-hcp (note the dash instead of underscore too, I think that is consistent with action/type naming elsewhere in the roles)

rawa-resul:
> Think the approach in general is sound, but needs some tweaking around the use of CLUSTER_TYPE ... use of the same env var across all roles needs to be consistent, so I would suggestion the logic here should be:
> 
> `if CLUSTER_TYPE=rosa-hcp`: , otherwise
> 
> Otherwise when CLUSTER_TYPE=roks/fyre/ipi etc we will have broken the `ocp_idms` role.
> 
> As proposed above, we will do the "normal" way unless it's rosa-hcp (note the dash instead of underscore too, I think that is consistent with action/type naming elsewhere in the roles)

I have changed the directory name to `rosa-hcp`. In the main.yml I check if cluster_type is set to rosa-hcp and run that otherwise I run the default task path directly (so that cluster_type is no longer part of the path).

ibm/mas_devops/roles/ocp_config/defaults/main.yml
```diff
@@ -1,4 +1,13 @@
 ---
+# Generic Settings
```
durera: Are these generic settings really generic?  I thought these were only used in rosa-hcp scenario from previous discussions?  e.g. should this maybe be:

```yaml
# ROSA-HCP Settings
# Important: cluster_type, cluster_name, rosa_token are only used when working with ROSA-HCP
# see <link> for more details
cluster_type: "{{ lookup('env', 'CLUSTER_TYPE') }}"
cluster_name: "{{ lookup('env', 'CLUSTER_NAME') }}"
rosa_token: "{{ lookup('env', 'ROSA_TOKEN') }}"
```

?

ibm/mas_devops/roles/ocp_config/tasks/update-additional-trusted-ca.yml
```diff
@@ -0,0 +1,116 @@
...
+      until:
+        - registry_additional_ca_cfgmap_lookup.resources is defined
+        - registry_additional_ca_cfgmap_lookup.resources | length > 0
+        - registry_additional_ca_cfgmap_lookup.resources[0].data is defined
+        - registry_additional_ca_cfgmap_lookup.resources[0].data | symmetric_difference(trusted_ca_dict) | length == 0
+      retries: 5
+      delay: 10
+
+    # 1.5. Cleanup
+    # Todo: delete the previous cfgmap in openshift-config
```
durera, line 111: Does this todo need addressing before we can merge this?

ibm/mas_devops/roles/ocp_idms/README.md
durera, line 1: While it's fresh in our head I think we should put a detailed explainer into this README explaining why and how ROSA-HCP is a special case .... everything that you think will be useful when/if someone comes along and needs to update this role and likely won't have any access to a development rosa-hcp cluster themselves.

ibm/mas_devops/roles/ocp_config/tasks/update-additional-trusted-ca.yml
rawa-resul: No just improvements to do in the future. Not necessary for the functioning of this.

ibm/mas_devops/roles/ocp_config/defaults/main.yml
rawa-resul: At this point only used for ROSA HCP. I guess if they become more generic we can move them out of there. I will get this updated.

ibm/mas_devops/roles/ocp_idms/README.md
rawa-resul: Sure, I will add that in. Btw I've also included a playbook that runs both the config and idms role for rosa hcp.

durera (approved):

PR #2402, opened 2026-08-06
[minor] update gpu cluster to latest channel and add cdi enabled

joseph122333:
### Description
This PR introduces updates to the NVIDIA GPU Operator deployment logic to support newer driver channels and enforce Container Device Interface (CDI) configurations. 

Specifically, this PR:
* **Updates the Default Channel:** Bumps the default `GPU_CHANNEL` from `v24.9` to `v26.3`.
* **Enables CDI:** Injects `cdi.enabled=true` and `cdi.default=true` into both the standard and custom-version `ClusterPolicy` templates to modernize device plugin handling.
* **Improves Idempotency & Updates:** Un-indented the `ClusterPolicy` deployment tasks out of the strict "only install if not present" block. This is a critical fix that ensures existing clusters will receive updates to their `ClusterPolicy` (like the new CDI fields) without requiring a manual teardown of the operator first.

### Test Results
- **Clean Installation:** Verified that the operator successfully deploys from scratch on an OpenShift cluster using the new `v26.3` channel.
- **Idempotency & Update Validation:** Ran the playbook against a cluster with the operator already installed. Confirmed that the `ClusterPolicy` tasks now properly evaluate the existing policy and apply the new `cdi` configuration gracefully.
- **Health Verification:** Confirmed that the `gpu-cluster-policy` instance successfully reached the `ready` state, and all downstream daemonsets (device-plugin, driver, dcgm-exporter, etc.) transitioned to `Running` seamlessly. 

### ⚠️ Notes for Reviewers
- Because the `ClusterPolicy` creation task was moved outside of the conditional check block, this will cause the updated policy (including the CDI changes) to automatically apply to all existing environments the next time their Ansible pipeline runs. Ensure you have understood the PR guidelines in the Playbook before proceeding with a review.


ibm/mas_devops/roles/nvidia_gpu/defaults/main.yml
```diff
@@ -3,7 +3,7 @@
 gpu_namespace: "{{ lookup('env', 'GPU_NAMESPACE') | default('nvidia-gpu-operator', true) }}"
 
 # Set channel for install and updates
-gpu_channel: "{{ lookup('env', 'GPU_CHANNEL') | default('v24.9', true) }}"
+gpu_channel: "{{ lookup('env', 'GPU_CHANNEL') | default('v26.3', true) }}"
```
dclain, line 6: let's update the default channel value in the README accordingly
https://github.com/ibm-mas/ansible-devops/blob/517a65a778fb01c77932451cafca84d10fcd5c34/ibm/mas_devops/roles/nvidia_gpu/README.md

dclain (commented):
too many changes in v2 template, we should only propagate the cdi update
README needs to be updated too

whitfiea (approved):

PR #2409, opened 2026-08-13
[patch] Fix MinIO listen address to work on non-IPv6 clusters

Divyesh-Khokhar:
## Description

MinIO was unconditionally using `[::]:9090` and `[::]:9000` as listen addresses, causing the route to fail on pure IPv4 clusters even though the pod was `1/1 Running`.

Updated the deployment to use `:9090` and `:9000` by default, which works on both IPv4 and IPv6 clusters. `[::]` is used only when `ocp_enable_ipv6=true`.

Also fixed the invalid `metadata` field in the container spec.

## Changes

* Use `:9090` and `:9000` by default.
* Use `[::]` only when `ocp_enable_ipv6=true`.
* Move the `tag` label to `spec.template.metadata.labels`.

### IPv4 Cluster

* MinIO pod: `1/1 Running`
* Route returned HTTP 200 and console loaded successfully.
* S3 connectivity verified from the aibroker pod. 

Screenshot 2026-08-13 at 3 56 45 PM
Screenshot 2026-08-13 at 3 57 01 PM
Screenshot 2026-08-13 at 3 57 16 PM
Screenshot 2026-08-13 at 3 58 26 PM

### IPv6 Cluster

* `ocp_enable_ipv6=true` continues to use `[::]:9090` and `[::]:9000`.

### DIFF
--- a/ibm/mas_devops/roles/minio/templates/minio/minio-deployment.yml.j2
+++ b/ibm/mas_devops/roles/minio/templates/minio/minio-deployment.yml.j2
@@ -12,6 +12,7 @@ spec:
     metadata:
       labels:
         app: "{{ minio_instance_name }}"
+        tag: "{{ minio_image_tag }}"
     spec:
       volumes:
         - name: storage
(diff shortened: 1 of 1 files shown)


karol-czarnecki (approved):
lgtm

jasmin-30 (approved):

terc1997 (approved):
Make sure to remove the all the address when removing Minio, but LGTM.

PR #2411, opened 2026-08-17
[minor] Add OLM/Helm-based installation support for CPD 5.3.1 

nehasangwai-ibm:
## Description

Adds Helm-based installation support for Cloud Pak for Data (CPD) 5.3.1+, alongside the existing OLM installation path. The `cp4d_service` role now auto-detects the appropriate installation method based on the CPD version and a component support matrix

### New features
- **Dual installation path**: `cp4d_service` now routes to either OLM (`tasks/olm/install-olm.yml`) or Helm (`tasks/helm/install-helm.yml`) based on `cpd_product_version` and the component support matrix
- **Component support matrix** (`vars/component-support.yml`): declarative mapping of which components (wsl, wml, ca, spark) support OLM and/or Helm installation
- **Helm repo setup** (`tasks/helm/setup-helm-repo.yml`): configures IBM charts Helm repo; airgap-aware — auto-detects registry host/port/credentials from cluster IDMS and pull-secret when `CPD_AIRGAP_HELM_REPO_URL` is not set
- **Helm component install** (`tasks/helm/install-component.yml`): installs CPD service Helm charts with dependency handling and retry logic
- **OLM-to-Helm migration** (`tasks/helm/cleanup-olm-resources.yml`): patches existing OLM-managed resources with Helm adoption labels when upgrading from CPD 5.2.0 → 5.3.1+; uses `vars/olm-resource-mappings.yml` for resource name mappings
- **OLM uninstall cleanup** (`roles/cp4d_uninstall/tasks/cleanup-helm-cluster-resources.yml`): removes Helm cluster-scoped CRDs, ClusterRoles, and ClusterRoleBindings left behind after namespace deletion; controlled by new `cpd_uninstall_delete_helm_cluster_resources` flag
- **Helm chart mirroring** (`roles/mirror_images/tasks/mirror_helm.yml`): mirrors Helm charts from CASE bundle to a target OCI registry for airgap deployments; supports direct, to-filesystem, and from-filesystem mirror modes

### Configuration changes
- New `defaults/main.yml` variables:
  - `cpd_install_method` (`auto`/`olm`/`helm`) — defaults to `auto`
  - `cpd_helm_repo_url`, `cpd_helm_repo_name`, `cpd_helm_timeout`, `cpd_helm_wait_retries`, `cpd_helm_wait_delay`
  - `cpd_airgap_helm_repo_url`, `cpd_helm_airgap` (airgap Helm registry detection)
  - `cpd_entitlement_key` / `ibm_entitlement_key` — support standalone role execution without parent `cp4d` role
  - Storage class fallback: `cpd_service_storage_class` and `cpd_service_block_storage_class` now fall back to `CPD_PRIMARY_STORAGE_CLASS` / `CPD_METADATA_STORAGE_CLASS`
- **CPD 5.3.1 added** to `common_vars/cp4d_supported_versions.yml` with `cpd_531` version flag
- New catalog source template: `roles/cp4d/templates/catalog_sources/5.3.1.yml`
- New olm-utils config map template: `templates/config_maps/olm-utils-cm-5.3.1.yml.j2`

## Test Results

epic - https://jsw.ibm.com/browse/MASR-9747
Dashboard - https://dashboard.ibmmas.com/tests/pfvtcpd & 

- [Y] CPD 5.3.1 Helm install (new path)
- [Y] CPD 5.2.0 → 5.3.1 upgrade (OLM-to-Helm migration via `cleanup-olm-resources.yml`)
- [Y] `mirror_images` Helm chart mirroring (direct, to-filesystem, from-filesystem modes)

<hr/>


ibm/mas_devops/playbooks/mirror_cli.yml
```diff
@@ -17,10 +17,8 @@
...
         extras_name: cli
         extras_version: "{{ cli_version }}"
-
-    - role: ibm.mas_devops.mirror_images
-      when: mirror_cli
-      vars:
-        manifest_name: extras_cli
-        manifest_version: "{{ cli_version }}"
-
+- role: ibm.mas_devops.mirror_images
```
durera: This doesn't seem like an intentional change/invalid yaml?  Am I missing something here, is there a reason to unindent this section?

ibm/mas_devops/roles/aiservice/tasks/config_database/main.yml
durera, line 1: Why are we deleting the white space between sections here?

ibm/mas_devops/roles/cp4d_service/defaults/main.yml
```diff
@@ -10,9 +10,44 @@
 # model of the Cloud Pak for Data team :/
 cpd_product_version: "{{ lookup('env', 'CPD_PRODUCT_VERSION') }}"
 
+# Installation Method Selection
+# -----------------------------------------------------------------------------
+# Auto-detect based on component support, or explicitly set
+cpd_install_method: "{{ lookup('env', 'CPD_INSTALL_METHOD') | default('auto', true) }}"
```
durera: What is the use case for anything other than auto?  If we support installing the same version of CPD in two different modes won't that create support issues going forward where we will have to cater to potentially two different types of install - our aim is consistency across installs.  Two installs of MAS + deps with the same catalog should be identical, with only configuration differences being at the application level, rather than how they were installed)

ibm/mas_devops/roles/cp4d_service/tasks/helm/install-component.yml
durera, line 1: General comment ... we should not be using `command` to perform help operations ... ansible has it's own `kubernetes.core.helm` that provides ansible-native helm support:

e.g. from the `longhorn` role:

```
# Install Longhorn
- name: "Install longhorn helm chart to {{ longhorn_namespace }}"
  kubernetes.core.helm:
    name: longhorn
    chart_ref: longhorn/longhorn
    release_namespace: "{{ longhorn_namespace }}"
    values: "{{ lookup('template', 'values.yml.j2') | from_yaml }}"
```

Due to time constraints we will merge as-is to ensure we can get CPD 5.3 support out the door ASAP, but we need to get an issue opened to address this as technical debt in a future update.

ibm/mas_devops/roles/cp4d_service/tasks/helm/install-helm.yml
```diff
@@ -0,0 +1,301 @@
...
+# This is the main entry point for Helm-based installations
+
+- name: "Setup Helm repository"
+  include_tasks: "setup-helm-repo.yml"
+
+# Only create ibm-entitlement-key if it does not already exist in the namespace.
+# The cp4d role is responsible for creating this secret with a valid password.
+# Writing it here unconditionally risks overwriting a valid secret with an empty
+# password when ibm_entitlement_key is not set (e.g. during mas update flows).
+- name: "Check if ibm-entitlement-key already exists in operator namespace"
```
durera, line 12: `ibm_entitlement_key` should never be used, `cpd_entitlement_key` only (which falls back to ibm_entitlement_key if necesseary) ... otherwise the updates to this role do not work if `CPD_ENTITLEMENT_KEY` is set.

Note how they are set up in `defaults/main.yml` (correct):
```
# IBM Entitlement Key Configuration
# These variables are defined in the parent cp4d role, but we need them here
# to support standalone execution of cp4d_service role (especially for Helm installations)
ibm_entitlement_key: "{{ lookup('env', 'IBM_ENTITLEMENT_KEY') }}"
cpd_entitlement_key: "{{ lookup('env', 'CPD_ENTITLEMENT_KEY') | default(ibm_entitlement_key, true) }}"
```
... but for this to work, all usage in the role needs to be via the latter

CPD_5.3.1_INSTALLATION_GUIDE.md
durera, line 1: This file should not be committed ... the content here should be reviewed and condolidted into the README(s) in the cpd and cpd service roles.

CPD_5.3.1_VERSION_MAPPING.md
durera, line 1: As with the install guide, this captures a lot of good information, but this needs to be reviewed and curated by a human (not AI) to extract the valuable information into the README files in cpd and cpd service roles

ibm/mas_devops/roles/cp4d_service/tasks/helm/install-component.yml
```diff
@@ -0,0 +1,179 @@
...
+        imagePullPrefix: cp.icr.io
+
+      {{ comp_assembly }}:
+        scaleConfig: {{ cpd_service_scale_config }}
+        version: {{ cpd_product_version }}
+        non_olm: true
+    dest: "/tmp/{{ component_name }}-helm-values.yaml"
+
+# Install cluster-scoped chart (CRDs, ClusterRoles, etc.)
+# Use helm template + oc apply for cluster-scoped to avoid Helm state management issues
```
durera: I think we need more explanation here ... what issues are we avoiding.  This seems like a hack/workaround for something -- but what?

I can see this is tied to ibm/mas_devops/roles/cp4d_uninstall/tasks/cleanup-helm-cluster-resources.yml .. I'm not saying this is wrong/there isn't a good reason to do it this way ... it's just that in the absence of any details it _seems_ the wrong way to be installing a helm chart, so we should explain why we are doing it for future reference and to enable better review/might result in an alternative solution coming up.

ibm/mas_devops/roles/cp4d_service/tasks/helm/install-component.yml
```diff
@@ -0,0 +1,179 @@
...
+        fileStorageClass: {{ cpd_service_storage_class }}
+
+        imagePullSecret: {{ ibm_entitlement_key_secret }}
+        imagePullPrefix: cp.icr.io
+
+      {{ comp_assembly }}:
+        scaleConfig: {{ cpd_service_scale_config }}
+        version: {{ cpd_product_version }}
+        non_olm: true
+    dest: "/tmp/{{ component_name }}-helm-values.yaml"
```
durera: We need to add a step at end of the task that deletes this file ... if there is no way to avoid using a temporary file then because we are storing credentials (`ibm_entitlement_key_secret`) to this file we need to ensure it's 100% deleted before the task completes.  This is an example where using the helm ansible module helps a lot too, e.g:

```
- name: "Install longhorn helm chart to {{ longhorn_namespace }}"
  kubernetes.core.helm:
    name: longhorn
    chart_ref: longhorn/longhorn
    release_namespace: "{{ longhorn_namespace }}"
    values: "{{ lookup('template', 'values.yml.j2') | from_yaml }}"
```

We don't need to manage the temporary helm values file in that role

ibm/mas_devops/roles/cp4d_service/tasks/helm/cleanup-olm-resources.yml
```diff
@@ -0,0 +1,163 @@
...
+    - olm_sa_check.resources is not defined or olm_sa_check.resources | length == 0
+  debug:
+    msg: "No existing OLM installation detected, proceeding with Helm installation"
+
+# Delete component CRs so Helm can recreate them with correct version
+# Only deletes CRs explicitly listed in component_cr_mapping
+- name: "Delete component CRs for {{ component_name }}"
+  when: component_name in component_cr_mapping
+  block:
+    - name: "Protect PVCs from cascade deletion"
```
durera: We can't directly interact with kubernetes via the `oc` or `kubectl` via ansible `shell` tasks.  

In certain user environments the context that `oc` will connect to is not the same context that ansible is connected to.

To write "scripted kubernetes stuff" we need to create custom ansible modules that use ansible's context for the connection ... there are a few examples in this repository already, e.g. see the scripts used by ocp_verify.

durera: Additionally, is this safe?

This cleanup is running in the content of a single cpd service, but it blats every PVC in the namespace.  If I run this for 4 services I'm doing this to all pvcs for all services 4 times ... potentially in parallel, potentially sequentially.

ibm/mas_devops/roles/cp4d_service/tasks/helm/install-component.yml
```diff
@@ -0,0 +1,179 @@
...
+
+# Airgap: the registry may store charts with a build-stamped version such as
+# "12.1.0+20260206.151111.23" while case_version is only "12.1.0".
+# Pull Chart.yaml without specifying --version so we get the actual tag pushed
+# to the registry, then use that full version string for helm template/upgrade.
+- name: "Resolve chart version from OCI registry (airgap)"
+  when: cpd_helm_airgap
+  shell: >
+    helm show chart {{ cpd_airgap_helm_repo_url }}/{{ comp_chart_name }}
+    --insecure-skip-tls-verify 2>/dev/null
```
durera: `--insecure-skip-tls-verify` ... feels like something that should be controlled by an env var, rather than on by default.  Otherwise we will get flagged for security problems here (and all the other places it's on by default).

durera (changes requested):

ibm/mas_devops/roles/cp4d_service/tasks/helm/install-component.yml
SahilSheikh1: Pushed changes, 
playbook will be reading "HELM_REGISTRY_CA_FILE" env. If set then it will pass it accordingly.

ibm/mas_devops/playbooks/mirror_cli.yml
nehasangwai-ibm: You're right, this was an accidental change introduced in this PR. It's now restored to match master

ibm/mas_devops/roles/aiservice/tasks/config_database/main.yml
nehasangwai-ibm: You're right, this wasn't intentional — the blank lines between sections were accidentally removed, Reverted to match master.

ibm/mas_devops/roles/cp4d_service/defaults/main.yml
nehasangwai-ibm: cpd_product_version (< 5.3.1 → OLM, >= 5.3.1 → Helm) and should never be a user choice. Having an explicit override risks two installs of the same CPD version ending up installed differently, which creates inconsistency and support issues.

Removed CPD_INSTALL_METHOD / cpd_install_method entirely and simplified install.yml to a single version-based resolve. No functional change — all existing installs behave identically.

ibm/mas_devops/roles/cp4d_service/tasks/helm/cleanup-olm-resources.yml
nehasangwai-ibm: Good catch. Replaced all command: oc label, command: oc annotate, and shell: oc patch/shell: oc delete tasks with kubernetes.core.k8s and kubernetes.core.k8s_info module calls so they use Ansible's connection context consistently.

You're right, it was unsafe. The original code fetched all PVCs in the namespace and removed ownerReferences from all of them — running for each of the 4 services meant touching every PVC up to 4 times, with a potential race condition if run in parallel.

Fixed by:

Fetching PVCs once per component invocation
Filtering to only PVCs whose ownerReferences point to this component's CR (matched by CR name for named CRs, by CR kind for name: all components like ws_runtimes)
So wsl only touches PVCs owned by ws-cr, wml only touches PVCs owned by wml-cr, etc. — no cross-component interference

ibm/mas_devops/roles/cp4d_service/tasks/helm/install-component.yml
nehasangwai-ibm: sure

ibm/mas_devops/roles/cp4d_service/tasks/helm/install-component.yml
nehasangwai-ibm: Temp file removed — replaced the copy task writing to /tmp with a set_fact that builds helm_values as a dict. The namespace-scoped chart now uses kubernetes.core.helm with values: "{{ helm_values }}" — no file written to disk, no credentials exposed.

oc removed from cluster-scoped chart — kubernetes.core.helm only supports upgrade --install so can't be used for helm template | oc apply. Instead, helm template now renders the manifests into a registered variable, which is then applied via kubernetes.core.k8s: definition: "{{ cluster_chart_manifests.stdout | from_yaml_all | list }}" — uses Ansible's connection context, no oc binary required.

ibm/mas_devops/roles/cp4d_service/tasks/helm/install-helm.yml
nehasangwai-ibm: Fixed — replaced ibm_entitlement_key with cpd_entitlement_key throughout. cpd_entitlement_key falls back to ibm_entitlement_key if not set, so both CPD_ENTITLEMENT_KEY and IBM_ENTITLEMENT_KEY are now handled correctly.

CPD_5.3.1_INSTALLATION_GUIDE.md
nehasangwai-ibm: sure, removing it

CPD_5.3.1_VERSION_MAPPING.md
nehasangwai-ibm: Removing it for now

ibm/mas_devops/roles/cp4d_service/tasks/helm/install-component.yml
nehasangwai-ibm: this is now stale — updated to use helm template → kubernetes.core.k8s (no oc, no temp file) following earlier review feedback.

durera (approved):

PR #2420, opened 2026-08-24
[minor] Switch Atlas alerts to webhook notifications with idempotency

Raahithyaj:
MASCORE-15204

## Description
Switched Atlas alert notifications from email to webhook (OCM). Webhook credentials are fetched at runtime from AWS Secrets Manager. Added idempotency to both project-level and cluster-level alert configuration, existing alerts are detected and skipped to prevent duplicates on re-runs.

## Test Results
Ran `ansible-playbook ibm.mas_devops.run_role` with `mongodb_action=install` against test Atlas project .

All 11 alert creation tasks correctly skipped — existing alerts (created via Terraform with both EMAIL + WEBHOOK notifications) were detected and no duplicates were created.

Screenshots attached for project-level (6 alerts) and cluster-level (5 alerts) showing all tasks skipping.
image
image

JIRA Tickets:
https://jsw.ibm.com/browse/MASCORE-15204
https://jsw.ibm.com/browse/MASCORE-16944


ibm/mas_devops/roles/mongodb/tasks/providers/atlas/configure_alerts_cluster.yml
```diff
@@ -17,11 +17,55 @@
...
+          --user "{{ atlas_public_key }}:{{ atlas_private_key }}" \
+          --silent --show-error --fail \
+          "{{ atlas_api_base_url }}/groups/{{ atlas_project_id }}/alertConfigs?pageNum={{ item }}&itemsPerPage=100"
+      register: atlas_alert_page_raw
+      changed_when: false
+      failed_when: atlas_alert_page_raw.rc != 0
+      no_log: true
+      loop: "{{ range(1, 1000) | list }}"
+      loop_control:
+        label: "page {{ item }}"
```
amitpandey0217: Can we use proper implementation using until for dynamic pagination instead of hardcoding? @mnivedithaa your thaught on this.

mnivedithaa: Yes. Can try to fetch total count and page accordingly

ibm/mas_devops/roles/mongodb/tasks/providers/atlas/configure_alerts_project.yml
```diff
@@ -32,15 +71,60 @@
...
     - name: "Configure basic performance alerts"
       block:
+        - name: "Check if CPU warning alert already exists"
+          set_fact:
+            cpu_warning_exists: >-
+              {{
+                atlas_all_alert_configs
+                | selectattr('eventTypeName', 'equalto', 'OUTSIDE_METRIC_THRESHOLD') | list
+                | selectattr('metricThreshold.metricName', 'equalto', 'NORMALIZED_SYSTEM_CPU_USER') | list
+                | selectattr('metricThreshold.threshold', 'equalto', atlas_alert_cpu_thresholds[0] | float) | list
```
mnivedithaa: Don't check threshold value. If we want to modify threshold value, we need to overwrite the existing alert only. Same review for other checks as well

ibm/mas_devops/roles/mongodb/tasks/providers/atlas/configure_alerts_project.yml
```diff
@@ -4,26 +4,65 @@
...
+        atlas_webhook_credentials: "{{ atlas_webhook_secret_output.stdout | from_json }}"
+      no_log: true
+
+    - name: "Set webhook url and secret from AWS Secrets Manager"
       set_fact:
-        atlas_alert_emails_list: "{{ atlas_alert_notification_emails if atlas_alert_notification_emails is iterable and atlas_alert_notification_emails is not string else (atlas_alert_notification_emails.split(',') if atlas_alert_notification_emails is string else atlas_alert_notification_emails | from_json) }}"
+        atlas_alert_webhook_url: "{{ atlas_webhook_credentials.webhook_url }}"
+        atlas_alert_webhook_secret: "{{ atlas_webhook_credentials.webhook_secret }}"
+      no_log: true
+
```
mnivedithaa, line 40: Add a validation step to ensure webhook_url and webhook_secret values are not empty

ibm/mas_devops/roles/mongodb/tasks/providers/atlas/configure_alerts_cluster.yml
Raahithyaj: The until: empty page condition already handles dynamic pagination, the loop stops as soon as a page returns empty results, so it never actually iterates 1000 times. The upper bound of 1000 is just a safety cap to prevent an infinite loop in Ansible.

ibm/mas_devops/roles/mongodb/tasks/providers/atlas/configure_alerts_project.yml
Raahithyaj: Makes sense, removing the threshold from the existence check prevents duplicate creation when the threshold value changes. However, this does mean that if the threshold is updated via the variable, the existing alert won't be automatically updated to reflect the new value. That would require a separate PUT to the existing alert's ID right?

ibm/mas_devops/roles/mongodb/tasks/providers/atlas/configure_alerts_project.yml
Raahithyaj: Added an assert after fetching webhook credentials from Secrets Manager to validate that both webhook_url and webhook_secret are non-empty.

ibm/mas_devops/roles/mongodb/tasks/providers/atlas/configure_alerts_cluster.yml
amitpandey0217: All good. Your point is valdi here.

amitpandey0217 (dismissed):

ibm/mas_devops/roles/mongodb/tasks/providers/atlas/configure_alerts_project.yml
Raahithyaj: image

After fixing the threshold value part ( to let it update the value if there is only the threshold value change and not comparing the threshold value for duplication check), this is the output.

amitpandey0217 (dismissed):

ibm/mas_devops/roles/mongodb/tasks/providers/atlas/configure_alerts_cluster.yml
```diff
@@ -17,11 +17,57 @@
...
+    - name: "Set total alert config count"
+      set_fact:
+        atlas_alert_total_count: "{{ (atlas_alert_count_raw.stdout | from_json).get('totalCount', 0) | int }}"
+
+    - name: "Fetch all existing alert configs from Atlas"
+      shell: |
+        curl --digest \
+          --user "{{ atlas_public_key }}:{{ atlas_private_key }}" \
+          --silent --show-error --fail \
+          "{{ atlas_api_base_url }}/groups/{{ atlas_project_id }}/alertConfigs?pageNum=1&itemsPerPage={{ [atlas_alert_total_count | int, 1] | max }}"
```
mnivedithaa, line 43: This is retrieving only 1 page with total count. Will this work if the total count exceeds 300?

ibm/mas_devops/roles/mongodb/tasks/providers/atlas/configure_alerts_project.yml
```diff
@@ -32,15 +80,65 @@
...
+    - name: "Set total alert config count"
+      set_fact:
+        atlas_alert_total_count: "{{ (atlas_alert_count_raw.stdout | from_json).get('totalCount', 0) | int }}"
+
+    - name: "Fetch all existing alert configs from Atlas"
+      shell: |
+        curl --digest \
+          --user "{{ atlas_public_key }}:{{ atlas_private_key }}" \
+          --silent --show-error --fail \
+          "{{ atlas_api_base_url }}/groups/{{ atlas_project_id }}/alertConfigs?pageNum=1&itemsPerPage={{ [atlas_alert_total_count | int, 1] | max }}"
```
mnivedithaa, line 111: same here

temp.py
```diff
@@ -0,0 +1,219 @@
+import subprocess, json, sys
+
```
mnivedithaa: Is this file required?

Raahithyaj: Oh this wasn't supposed to be pushed,I used this to test the logic before porting to the main files and   I was supposed to delete it before committing.

ibm/mas_devops/roles/mongodb/tasks/providers/atlas/configure_alerts_cluster.yml
Raahithyaj: In the Atlas api documentation [here](https://www.mongodb.com/docs/api/doc/atlas-admin-api-v2/operation/operation-listgroupalertconfigs), they have mentioned the upper limit as 500, so anything below that should be fetch-able. 
Earlier It wasn't throwing error beyond 300 number due to the presence of a faulty alert in that range. when some alerts were deleted that alert fell in lower range and the error came within 200 range.

mnivedithaa (approved):

whitfiea (approved):

